PR #2, opened 2019-02-01
Added first geodesics solver

ritzvik:


shreyasbapat:
I would suggest you to configure your  local it by:
```
git config --global user.name "Ritwik Saha"
git config --global user.email jobhiemail@jobhi.com
```

PR #3, opened 2019-02-02
Added seperate class for solving ODEs and Fixed scaling factors

ritzvik:


src/einsteinpy/OdeMethods/RK4thOrder.py
```diff
@@ -0,0 +1,45 @@
+import numpy as np
```
shreyasbapat: Can we not name the module will all small letters?

src/einsteinpy/OdeMethods/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .RK4thOrder import RK4thOrder
```
shreyasbapat: A module name is always small.

src/einsteinpy/OdeMethods/__init__.py
```diff
@@ -0,0 +1,6 @@
...
+
+try:
+    from scipy.integrate import RK45 as RK45Scipy
```
shreyasbapat: Why this extra piece?

PR #33, opened 2019-02-06
Add travis once in for all

shreyasbapat:


shreyasbapat:
@Masquerade0097 @ritzvik @spino17  I would expect a green tick like this on every PR we make

PR #51, opened 2019-02-10
Add config for codecov

shreyasbapat:


shreyasbapat:
I am merging this even if it is failing! As it was earlier considering tests in the coverage as well. Now it doesn't

PR #56, opened 2019-02-13
Added tests for constant

shreyasbapat:


shreyasbapat:
I think 0.06% is a good sign! Merging! Didn't expect such a rise in coverage

PR #81, opened 2019-03-02
[WIP] Add kerr null-geodesics

ritzvik:
Referring to the papers
https://arxiv.org/pdf/1601.02063.pdf
http://www.math.mcgill.ca/gantumur/math599w17/project-kerr.pdf

### DIFF
--- a/src/einsteinpy/metric/kerr_null.py
+++ b/src/einsteinpy/metric/kerr_null.py
@@ -145,7 +145,7 @@ class KerrNull:
             + (self.a ** 2 * E ** 2)
         )
 
-    def f_vec(self, sixvec):
+    def f_vec(self, ld, sixvec):
         sigma = self.sigma(sixvec[1], sixvec[2], self.a)
         delta = self.delta(sixvec[1], self.schwarzschild_r, self.a)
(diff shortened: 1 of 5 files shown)


src/einsteinpy/metric/kerr_null.py
```diff
@@ -0,0 +1,83 @@
...
+# paper : https://arxiv.org/pdf/1601.02063.pdf
+# paper : http://www.math.mcgill.ca/gantumur/math599w17/project-kerr.pdf
+
+_c = constant.c.value
+_G = constant.G.value
+
+
+class KerrNull:
+    """
+    Class for defining methods for Kerr Null Geodesics (Schwarzschild Null Geodesics when J=0)
```
shreyasbapat: My proposal, how about we define a class explicitly, which inherits this class. Which is Schwarzschild Null?

ritzvik: Yes. It's a good idea.
But let's first write KerrNull

shreyasbapat:
Reminder: Just 2 Days for release of 0.1.0

ritzvik:
Some related repo : 
https://github.com/grtrace/grtrace

PR #100, opened 2019-03-08
New symbolic feature + some essential fixes

ritzvik:


shreyasbapat:
Do you want this to be backported?

src/einsteinpy/utils/christoffel.py
```diff
@@ -80,6 +80,36 @@ def christoffels(list2d, syms):
     return christlist
 
 
+def simplify_tensor_helper(v):
```
shreyasbapat: Make this private!

src/einsteinpy/utils/christoffel.py
```diff
@@ -80,6 +80,36 @@ def christoffels(list2d, syms):
...
+    ndlist : list
+        N-Dimensional list of sympy expressions representing some tensor
+    
+    Returns
+    -------
+    list
+        N-Dimensional list
+    
+    """
+    return simplify_tensor_helper(ndlist)
```
shreyasbapat: Here too!

src/einsteinpy/metric/schwarzschild.py
```diff
@@ -4,10 +4,9 @@
 import numpy as np
 
 from einsteinpy import constant
-from einsteinpy.integrators import RK45, RK4naive
```
shreyasbapat: Great!

src/einsteinpy/metric/schwarzschild.py
```diff
@@ -110,7 +109,7 @@ def from_cartesian(cls, pos_vec, vel_vec, time, M):
     def christ_sym1_00(self, vec):
         num1 = (-2 * _G * self.M.value) + ((_c ** 2) * vec[1])
         num2 = _G * self.M.value
-        deno1 = _c ** 4
+        deno1 = _c ** 2
```
shreyasbapat: Change in formula?

src/einsteinpy/utils/christoffel.py
shreyasbapat: Also, the parameter, should be more than `v`

ritzvik: > Also, the parameter, should be more than `v`

actually the parameter can be both sympy expression or list. So better keep it ```variable``` or ```v```

src/einsteinpy/metric/schwarzschild.py
ritzvik: > Change in formula?

The initial sources from which we got metric was faulty. I got the insight while developing Kerr Metric. That's why all scaling factors were needed to be introduced

ritzvik:
> Do you want this to be backported?

What do you mean by this?

shreyasbapat:
Do you want this change in 0.1.1 or directly 0.2.0?

ritzvik:
> Do you want this change in 0.1.1 or directly 0.2.0?

0.1.1 would be good

PR #155, opened 2019-04-07
Fix Links throughout

shreyasbapat:
Fixes Nothing

Add einsteinpy.org everywhere!


ritzvik:
Einsteinpy main website still redirects to RTD page when clicked on documentation. I believe this has to fixed from repo `einsteinpy.gihtub.io`

PR #194, opened 2019-04-12
 Reorder parameters within functions

### LINKED ISSUE #141: Organize and reorder utils for Kerr, Kerr-Newman and Schwarzschild metrics
🐞 **Problem**
As of now, various useful functions used by `metric/schwarzschild.py` are in `utils/schwarzschild_utils.py`. Same goes for `metric/kerr.py` and `metric/kerrnewman.py`. We need to move these functions out of `utils` and make them a core part of library as they are quite-useful even when used stand-alone.

🎯 **Goal**

  - Migrate them somewhere, so as to make them look like a core part of library.
  - Reorder the parameters inside the functions, so as parameters like `c`, `G` etc are made to be optional with default values can be taken in SI units from `constant.py`.
  - Many functions take `Rs` as parameters which is Schwarzschild Radius. However, there is no sense to keep parameters `Rs` as it can be calculated from more fundamental things like mass, speed of light(`c`), gravitational constant(`G`). `Rs = 2*M*G/square(c)`.
  - These changes would break the API, and therefore requires updation of tests and example notebooks.

⏫ **Progress**

The tasks to be taken up in separate PRs
  - [x] Reorder parameters within functions
  - [ ] Move the functions somewhere more suitable

💡 **Possible solutions**

  - Firstly, wherever the functions take `Rs`, i.e schwarzschild radius as an input, replace them with `M`,`G`, and `c` and calculate `Rs = 2*G*M/(c**2)` inside the function itself. We can't expect the user to supply schwarzschild radius. While implementing it, leave out function `delta()` in both `kerr_utils.py` and `kerrnewman_utils.py`.

  - Currently functions are like `func(c, G, Cc, r, theta.....)`. Here the fundamental constants need to be optional arguments and therefore, should be last few parameters like `func(r, theta, ..... , c=default_value, G=default_value, Cc=default_value)`. Leave out `G` and `Cc` whenever not required.
    - `c` is Speed of light. Default value : `einsteinpy.constant.c.value`
    - `G` is gravitational constant. Default value : `einsteinpy.constant.G.value`
    - `Cc` is coulomb's  constant. Default value : `einsteinpy.constant.coulombs_const.value`
  - After the changes, the API would break, therefore, update the example notebooks.
  - Update the user guide
  - Update the tests

 * Submit a pull request and add this in comments - `Addresses #141`

alpeshjamgade:
Fixes #141

### DIFF
--- a/src/einsteinpy/metric/kerr.py
+++ b/src/einsteinpy/metric/kerr.py
@@ -117,7 +117,7 @@ class Kerr:
 
     def f_vec(self, ld, vec):
         chl = kerr_utils.christoffels(
-            _c, vec[1], vec[2], self.schwarzschild_r.value, self.a
+            vec[1], vec[2], self.M, self.a
         )
         vals = np.zeros(shape=(8,), dtype=float)
         for i in range(4):
@@ -195,7 +195,7 @@ class Kerr:
         )
         _scr = self.schwarzschild_r.value * 1.001
         _event_hor = (
-            kerr_utils.event_horizon(self.schwarzschild_r.value, self.a)[0] * 1.001
+            kerr_utils.event_horizon(self.M, self.a)[0] * 1.001
         )
         while ODE.t < end_lambda:
             vec_list.append(ODE.y)
@@ -269,7 +269,7 @@ class Kerr:
         )
         _scr = self.schwarzschild_r.value * 1.001
         _event_hor = (
-            kerr_utils.event_horizon(self.schwarzschild_r.value, self.a)[0] * 1.001
+            kerr_utils.event_horizon(self.M, self.a)[0] * 1.001
         )
 
         def yielder_func():
--- a/src/einsteinpy/metric/kerr.py
+++ b/src/einsteinpy/metric/kerr.py
@@ -116,9 +116,7 @@ class Kerr:
         return cls._classmethod_handler(bl_pos_vec, bl_vel_vec, time, M, a)
 
     def f_vec(self, ld, vec):
-        chl = kerr_utils.christoffels(
-            vec[1], vec[2], self.M, self.a
-        )
+        chl = kerr_utils.christoffels(vec[1], vec[2], self.M, self.a)
         vals = np.zeros(shape=(8,), dtype=float)
         for i in range(4):
             vals[i] = vec[i + 4]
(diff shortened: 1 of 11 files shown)


ritzvik:
Really great contribution @alpeshjamgade . Thanks

I would be reviewing in a day or two. Meanwhile could you make one change.
define a private function `_scr(M, c=default value, G=default value` for calculating `Rs`. Putting units for calculating `Rs` seems unnecessary and slow, and may cause problems while implementing `jit` for the functions.

Also, I guess you have made some merge commits while rebasing your branch. Try to fix that. Maybe refer https://github.com/k88hudson/git-flight-rules/ . 
One word of advice, try to create separate branches for fixing and developing things. The master should not ever diverge from the upstream. Believe me, I have learnt that the hard way.

alpeshjamgade:
> 
> 
> Really great contribution @alpeshjamgade . Thanks
> 
> I would be reviewing in a day or two. Meanwhile could you make one change.
> define a private function `_scr(M, c=default value, G=default value` for calculating `Rs`. Putting units for calculating `Rs` seems unnecessary and slow, and may cause problems while implementing `jit` for the functions.
> 
> Also, I guess you have made some merge commits while rebasing your branch. Try to fix that. Maybe refer https://github.com/k88hudson/git-flight-rules/ .
> One word of advice, try to create separate branches for fixing and developing things. The master should not ever diverge from the upstream. Believe me, I have learnt that the hard way.

thank you

you asked me to define a private function _scr. 
Before the schwarzschild_radius function in schwarzschid_utils.py  was getting any M value as input weather it has astropy unit or not  and return Rs in astropy units and also 'Rs' was the input to the most functions which was in astropy units only, but  when i changed Rs with "M", i needed to make sure that "M" should be in astropy units , that means everytime i call such function i need to provide M value in astropy units otherwise it would throw an error, and we dont expect users to provide M in astropy units, it would be great if they can just give M value as input and if the given function need to calculate Rs, it would need M in astropy units then schwarzschild_radius function will convert M to astropy units, thast just  what i did. please tell me if i can do any better.

alpeshjamgade:
i think we dont need private  _scr function, if we want Rs in floats we can just asked for it with command Rs.value , otherwise it will always be in astropy units. please tell me if i dont understand something.

ritzvik:
> i think we dont need private _scr function, if we want Rs in floats we can just asked for it with command Rs.value , otherwise it will always be in astropy units. please tell me if i dont understand something.

User always specifies things with units, but during heavy calculation internally, we don't use units. As functions like `christoffels()` are called thousands of time in a loop to get the trajectory it may be good to save some computation by not using `astropy.units`. But still wait some time, I am not sure what to do. @shreyasbapat Suggestions??

alpeshjamgade:
i have just created another pull request #196 for this issue , this time i didnt messed up my master. sorry for this. i did not consider on which branch i was making changes before.

PR #251, opened 2019-06-25
upadated coordinate conversion

### LINKED ISSUE #236: Update the User Guide to the newest version
🐞 **Problem**

https://docs.einsteinpy.org/en/latest/user_guide.html

This does not use the `CartesianDifferential` objects that we created a while ago.

Refer to : https://docs.einsteinpy.org/en/latest/examples/Visualizing_advance_of_perihelion_of_a_test_particle_in_Schwarzschild_space-time.html

for understanding the changes done.
🎯 **Goal**

To get the User Guide up and running! 

💡 **Possible solutions**

shilpi1958:
Fixes #236

@ritzvik 
@shreyasbapat 

Just a query.
In CartesianDifferential class, we do not have a function returning just the velocity. Shouldn't there be one?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,10 +49,6 @@ python_requires = >=3.5
 [options.packages.find]
 where = src
 
-[options.entry_points]
-console_scripts =
-    einsteinpy = einsteinpy.cli:main
-
 [options.extras_require]
 dev =
         numba>=0.39 ; implementation_name=="cpython"


ritzvik:
> Just a query.
> In CartesianDifferential class, we do not have a function returning just the velocity. Shouldn't there be one?

Yups, you are right! we haven't thought of it.

shilpi1958:
I would like to work on it then!

docs/source/user_guide.rst
```diff
@@ -114,17 +114,33 @@ Using the functions:
...
+        print(bl_pos)
+
+        cartsn_pos = bl_pos.to_cartesian(a)
+        print(cartsn_pos)
+
+        pos_vel_coord = CartesianDifferential(.265003774 * u.km, -153.000000e-03 * u.km,  0 * u.km,
+                                  145.45557 * u.km/u.s, 251.93643748389 * u.km/u.s, 0 * u.km/u.s)
+
+        bl_coord = pos_vel_coord.bl_differential(a)
+        bl_coord = np.array(bl_coord)
```
ritzvik: I don't think this line would work.
I guess what to you want is `bl_coord = bl_coord.si_values(); bl_vel = bl_coord[3:]`
Try them on your system before changing it here :)

docs/source/user_guide.rst
```diff
@@ -114,17 +114,33 @@ Using the functions:
...
+        bl_vel = bl_coord[3:]
+        print(bl_vel)
+
+        cartsn_coord = bl_coord.cartesian_differential(a)
+        cartsn_coord = np.array(cartsn_coord)
+        cartsn_vel = cartsn_coord[3:]
```
ritzvik: same here 😄

docs/source/user_guide.rst
shilpi1958: I tried both np.array and si_values
Both worked on my system

docs/source/user_guide.rst
```diff
@@ -151,7 +167,6 @@ EinsteinPy also supports smbolic calculations in :py:class:`~einsteinpy.utils.ch
 
 Future Plans
 ============
```
ritzvik: blank line was required here I guess

ritzvik: try to figure out why docs are not building https://circleci.com/gh/einsteinpy/einsteinpy/3278?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
I am a little preoccupied :)

shilpi1958: oh, okay!

ritzvik:
> I would like to work on it then!

Kindly go ahead!! Thanks 😃

docs/source/user_guide.rst
ritzvik: are you sure? lemme check!

ritzvik: gimme a screenshot. No method for this is present. How is this working??

shilpi1958: I tried it on a rough jupyter notebook.
And I made the changes you suggested . I thought maybe someone may not want to change the units that why I chose np.array 



Screen Shot 2019-06-25 at 2 56 59 PM

ritzvik: As I can see there is no difference b/w `bl_coord` and `np.array(bl_coord)`

ritzvik:
One small suggestion, kindly create new branches for development from next PR onwards.
master is meant to be in sync or may fall behind upstream, but it never diverges.
Learnt it the hard way!

docs/source/user_guide.rst
shilpi1958: you are right! I misunderstood

shilpi1958: I changed it to si_values()

ritzvik:
> I would like to work on it then!

Would you like to do it in the same PR??

shilpi1958:
> > I would like to work on it then!
> 
> Would you like to do it in the same PR??

yes

src/einsteinpy/coordinates/velocity.py
```diff
@@ -1,7 +1,7 @@
 import astropy.units as u
 import numpy as np
 
-from .core import BoyerLindquist, Cartesian, Spherical
```
ritzvik: removing the `.` wouldn't work. It's about absolute n relative imports

src/einsteinpy/coordinates/velocity.py
```diff
@@ -61,6 +61,18 @@ def si_values(self):
             )
         )
 
+    def vel_coord(self):
```
ritzvik: `def velocities(return_np = False):` can be a better name. When `False`, you can return a list `[self.v_x, self.v_y, self.v_z]`.
When `True`, you can return `self.si_values()[3:]`. How does it sound?

There is no point returning strings, maybe the user would want to do something with the numbers.

Also, we should try to avoid `_`s as much as possible while naming functions :D

shilpi1958: yeah that would be better!

ritzvik: for returning string, a separate function can be made. It's not quite a concern!!

shilpi1958: I created a new PR for velocity. Shall I revert the last commit?

src/einsteinpy/coordinates/velocity.py
shilpi1958: S what should I do here?

ritzvik: just revert to `from .core import ....`

src/einsteinpy/coordinates/velocity.py
```diff
@@ -61,6 +61,22 @@ def si_values(self):
             )
         )
 
+    def velocities(self):
```
ritzvik: `def velocities(self, return_np = False)`

src/einsteinpy/coordinates/velocity.py
ritzvik: no need. we'll clean this up!

src/einsteinpy/coordinates/velocity.py
shilpi1958: Then it doesn't work on my system. I used einsteinpy.coordinates then !

Issue #250, opened 2019-06-25
__str__ and __repr__ for Tensor Class


ritzvik:
🎯 **Goal**

  - Additions in `einsteinpy/symbolic/tensor.py`.
  - The heading of the issue is self-explanatory.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

OVSofia:
I'll start with this when I fix the mistakes of my last PR

Issue #70, opened 2019-02-28
Evolve symbolic calculations related to GR as a seperate module within the library


ritzvik:
🐞 **What bugs you?**
Along with being a library for numerical relativity, the library should support calculation of core GR related quantities in their symbolic form using [Sympy](https://www.sympy.org/). Currently, some symbolic functions reside in `christoffel.py` in `utils`.

🎯 **Goal**
  - Move functions to a separate module and make this a full fledged feature of the library.
  - Add many more useful quantities.
  - Make the library one-stop solution for GR problems.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #70`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

ritzvik:
@shreyasbapat add respective labels

ritzvik:
This looks promising :
  - https://github.com/spacetimeengineer/spacetimeengine

rishi-s8:
There’s already an example about some of these aspects in examples of SymPy. 
I think [this](https://github.com/sympy/sympy/blob/master/examples/advanced/relativity.py) would help.

Marlin-Na:
For reference:
1. [GraviPy](https://pypi.org/project/GraviPy/) has implemented several metrics, including Christoffel symbols, Ricci tensor, Riemann tensor, Einstein tensor and Geodesics.
2. [SageManifolds](https://sagemanifolds.obspm.fr/examples.html) in Sage.

ritzvik:
We can close this I think, as very specific symbolic related issue are being opened now.

Way to go @OVSofia

PR #261, opened 2019-06-28
Overhaul of symbolic modules and tensor algebra

### LINKED ISSUE #256: Implement index raising and lowering in Tensors
🐞 **Problem**
We have no way to know whether the Tensor has contravariant or covariant indices

🎯 **Goal**

The indices must be properly formatted in front of a symbol.

  - [x] Metric Tensor
  - [x] Christoffel Symbols
  - [x] Riemann Tensor
  - [x] Ricci Tensor
  - [x] Weyl Tensor

💡 **Possible solutions**

cjayross:
Fixes #256, #250, #70

The current implementation for symbolically representing tensors is simply a wrapper of the ~sympy.Array class. As such, it does not support symbolic subscripting and superscripting, tensor algebra, automatic tensor contraction and tensor products, automatic application of the metric to raise and lower indices, or lazy evaluation.

This pull requests integrates the module ~sympy.tensor.tensor into a format that is much more effective in doing abstract calculations with tensors having known values as arrays.

Example usage:
```python
>>> from sympy import diag, symbols
>>> from einsteinpy.symbolic.tensor import *
>>> E1, E2, E3, B1, B2, B3 = symbols('E1:4 B1:4')
>>> em = [[0, -E1, -E2, -E3],
          [E1, 0, -B3, B2],
          [E2, B3, 0, -B1],
          [E3, -B2, B1, 0]]
>>> eta = Metric('eta', diag(1, -1, -1, -1))
# Using the notation adopted by sympy, special symmetries can optionally
# be applied to tensors. Here, [[2]] denotes that the tensor is antisymmetric.
>>> F = Tensor('F', em, eta, symmetry=[[2]])
>>> alpha, beta = indices('alpha beta', eta)
>>> expr = F(alpha, beta) + F(beta, alpha)
>>> expand_tensor(expr)
0
>>> expr = F(alpha, beta) * F(-alpha, -beta)
# eta_0 and eta_1 are dummy indices denoting contraction.
>>> expr
F(eta_0, eta_1) * F(-eta_0, -eta_1)
# To see the algebraic result of the contraction, ``expand_tensor`` is used.
>>> expand_tensor(expr)
2*B_1**2 + 2*B_2**2 + 2*B_3**2 - 2*E_1**2 - 2*E_2**2 - 2*E_3**2

# The following examples shows the calculation of the squared magnitude
# of the energy-momentum in a Schwarzschild spacetime.
>>> r, th = symbols('r theta', real=True)
>>> E, p1, p2, p3 = symbols('E p_1:4', positive=True)
>>> schwarzschild = diag(1 - 1/r, -1/(1 - 1/r), -r**2, -r**2*sin(th)**2)
>>> g = Metric('g', schwarzschild)
>>> p = Tensor('p', [E, p1, p2, p3], g)
>>> mu, nu = indices('mu nu', g)
>>> expr = g(-mu,-nu) * p(mu) * p(nu)
>>> expand_tensor(expr)
E**2*(1 - 1/r) - p_1**2/(1 - 1/r) - p_2**2*r**2 - p_3**2*r**2*sin(theta)**2
# The same result would have been obtained by setting ``expr = p(mu) * p(-mu)``.
```

This code does not yet support differentiation however. As such it does not implement a method for calculating the Christoffel symbols, etc. My intentions are to see about representing derivatives as simply being non-commutative tensors themselves and do differentiation in ``expand_tensor``.

Another feature I hope to add is the ability to define partial indices that range through only a subset of the components. For example, to provide the option to define indices for only the spatial components of a metric on spacetime. This will be helpful for metric decomposition (such as when studying gravitational radiation).

Additionally, I also hope to add additional methods for applying coordinate transformations as well as reducing the ranks of tensors by applying constraints on coordinates (i.e. representing hypersurfaces with induced metrics).

Lastly, I would like to also provide an easy way to lambdify tensors so that their components are numpy-like functions rather than symbolic representations.


cjayross:
Currently, my efforts for supporting differentiation are beginning to show results, but there are still a number of bugs still needing to be sorted out.

cjayross:
I've gotten to the point by which the code involved in the pull request is ready to begin more formal testing.

Additions made include differentiation support and the means to calculate the Christoffel symbols, the Riemann curvature tensor, the Ricci tensor/scalar, and the Weyl conformal tensor.

shreyasbapat:
Hi @cjayross , Let's do one thing. Instead of replacing the symbolic module which @OVSofia is developing, is it okay to first create a `tensor` module. When both the modules are sufficiently developed, we can see what are the common things and then finally merge them into one? 

Also, for the quality checks, 
when in `einsteinpy` folder, run `tox -e reformat` and the quality checks will start passing :)

shreyasbapat:
I will thoroughly review the code once you feel you are done with it! Also, as it is a very big Pull request, almost 500-600 lines are there in diff, I will ask you to explain briefly every major addition!

ritzvik:
> Hi @cjayross , Let's do one thing. Instead of replacing the symbolic module which @OVSofia is developing, is it okay to first create a `tensor` module. When both the modules are sufficiently developed, we can see what are the common things and then finally merge them into one?
> 
> Also, for the quality checks,
> when in `einsteinpy` folder, run `tox -e reformat` and the quality checks will start passing :)

`tensor` wouldn't be an elegant nomenclature for the new module, it's too generic. Maybe `symGR` or let @cjayross suggest one. There is no hint of symbolic calculation in the name `tensor`.

cjayross:
@shreyasbapat I'll work on providing a detailed explanation soon. Until then, you can read the documentation in the code, however it's not the best I don't think.

src/einsteinpy/symbolic/tensor.py
```diff
@@ -1,55 +1,327 @@
...
-        ~sympy.tensor.array.dense_ndim_array.ImmutableDenseNDimArray
-            Sympy Array object
-        
         """
-        return self.arr
+        array = self.as_array()
+        for pos, idx in enumerate(indices):
+            if idx.is_up ^ (self.covar[pos] > 0):
+                if idx.is_up:
+                    metric = idx.tensor_index_type.metric.as_inverse()
```
cjayross: Since this index implies that the array needs to have the components represented by this index be raised, the inverse metric is used.

src/einsteinpy/symbolic/tensor.py
```diff
@@ -1,55 +1,327 @@
...
+                else:
+                    metric = idx.tensor_index_type.metric.as_array()
+                new = tensorcontraction(tensorproduct(metric, array), (1, 2 + pos))
+                permu = list(range(len(indices)))
+                permu[0], permu[pos] = permu[pos], permu[0]
+                array = permutedims(new, permu)
+        return array
 
     def simplify(self):
```
cjayross: This method calls ``~sympy.simplify`` on the stored array and replaces it with the result. This also updates the ReplacementManager to also map indexed tensors of this object to the simplified array. This method is kept separate because it's possible for ``simplify`` to be very slow.

src/einsteinpy/symbolic/tensor.py
```diff
@@ -1,55 +1,327 @@
...
 
         """
-        return sympy.simplify(self.tensor())
+        array = simplify(self.as_array())
+        self._array = array
+        ReplacementManager.replace(self, array)
+        return array
+
+
+class Index(TensorIndex):
```
cjayross: This class works exactly like a ``~sympy.Symbol`` object, except it carries with it a reference to a metric.

src/einsteinpy/symbolic/tensor.py
```diff
@@ -1,55 +1,327 @@
...
+    """
+
+    def __new__(cls, symbol, metric, is_up=True, **kwargs):
+        return super().__new__(cls, symbol, metric, is_up=is_up, **kwargs)
+
+    def __neg__(self):
+        return Index(self.name, self.tensor_index_type, (not self.is_up))
+
+
+def expand_tensor(expr, idxs=None):
```
cjayross: This is the function by which performs the calculations represented by a tensor expression. Whenever you are finished doing algebra symbolically, this is the method you use to get an array equivalent to the said expression.

src/einsteinpy/symbolic/tensor.py
```diff
@@ -1,55 +1,327 @@
...
+        a = [x.name for x in symbols(s, seq=True)]
+    else:
+        raise ValueError(f"expected a string, received object of type {type(s)}")
+    idxs = [Index(idx, metric, is_up) for idx in a]
+    if len(idxs) == 1:
+        return idxs[0]
+    return idxs
+
+
+# metric tensors and general tensors commute with each other and themselves.
```
cjayross: It's probably redundant to have these two commutation groups be separated. I'll probably remove the ``metric`` group.

src/einsteinpy/symbolic/metric.py
```diff
@@ -1,45 +1,248 @@
...
+            self._weyl = Tensor(
+                "C", res, self, symmetry=[[2, 2]], covar=(1, -1, -1, -1)
+            )
+        return self._weyl
+
+    @property
+    def einstein(self):
+        r"""
+        Returns the Einstein tensor using the formula:
+        G_{\mu\nu} = R_{\mu\nu} - \frac{1}{2} R g_{\mu\nu}
```
cjayross: equation

shreyasbapat:
Closed for  https://github.com/einsteinpy/riccipy/pull/1

PR #274, opened 2019-07-02
Add plotter classes for matplotlib and plotly.

Varunvaruns9:
Added basic Matplotlib and Plotly plotter classes with a Plotter class that automatically switches between the two depending on the platform used.

I'll add the animate function in another PR within a few days.


shreyasbapat:
Run `tox -e reformat`

Varunvaruns9:
`mypy` is giving an error because I am using `get_ipython` to check if we are using a notebook or not. It gives a not defined error which is what it is supposed to do if run on the command line. I am ignoring that. Is it ok?

src/einsteinpy/plotting/plotter.py
```diff
@@ -0,0 +1,70 @@
+import random
+
+from plotly.offline import init_notebook_mode
+
+from .mplplotter import MatplotlibPlotter
+from .plotlyplotter import PlotlyPlotter
+
+
+def in_ipynb():
```
shreyasbapat: Brilliant! 
Yes, you may skip that line :)

src/einsteinpy/plotting/plotter.py
```diff
@@ -0,0 +1,70 @@
...
+        if shell == "ZMQInteractiveShell":
+            return True
+        if shell == "TerminalInteractiveShell":
+            return False
+        return False
+    except NameError:
+        return False
+
+
+class Plotter(MatplotlibPlotter, PlotlyPlotter):
```
shreyasbapat: Maybe rename it to `GeodesicPlotter` ?

src/einsteinpy/plotting/plotter.py
```diff
@@ -0,0 +1,70 @@
...
+
+        """
+        if self._notebook:
+            PlotlyPlotter.plot(self, geodesic, color=color)
+        else:
+            MatplotlibPlotter.plot(self, geodesic, color=color)
+
+    def show(self):
+        if self._notebook:
+            return PlotlyPlotter.show(self)
```
shreyasbapat: No return here !

shreyasbapat: I would say, do this if-else only once! There's too much redundant code :)

src/einsteinpy/plotting/plotlyplotter.py
```diff
@@ -0,0 +1,71 @@
...
+            Geodesic of the body
+        color : hex code RGB, optional
+            Color of the dashed lines. Picks a random color by default.
+
+        """
+        vals = geodesic.trajectory
+        r = np.array([coord[1] for coord in vals])
+        phi = np.array([coord[3] for coord in vals])
+
+        x = r * np.cos(phi)
```
shreyasbapat: This part is wrong! As discussed in #277 
@ritzvik Can you quickly tell how to correct it?

src/einsteinpy/plotting/mplplotter.py
```diff
@@ -0,0 +1,89 @@
+import random
+
+import matplotlib.pyplot as plt
+import numpy as np
+from matplotlib.patches import Circle
+
+
+class MatplotlibPlotter:
```
shreyasbapat: Maybe name the class as StaticGeodesicPlotter?  :P

src/einsteinpy/plotting/plotlyplotter.py
```diff
@@ -0,0 +1,71 @@
+import os
+import random
+
+import numpy as np
+import plotly.graph_objs as go
+from plotly.offline import plot as saveplot
+
+
+class PlotlyPlotter:
```
shreyasbapat: InteractiveGeodesicPlotter ?

src/einsteinpy/plotting/mplplotter.py
Varunvaruns9: Should I remove the old ones then?

shreyasbapat: No! Put them in a separate folder. Name that folder, "senile" :P

shreyasbapat: And anyways, put this all in geodesics folder :)

Varunvaruns9:
I changed the `GeodesicPlotter` class to work without any extra code.

@ritzvik I added `return_cartesian=True` to `calculate_trajectory()` but it seems to have slowed down creating a new object of class `Geodesic` significantly. Maybe the method can be optimized to work faster?

src/einsteinpy/plotting/geodesics/__init__.py
```diff
@@ -1,2 +1,5 @@
-from einsteinpy.plotting.geodesics.geodesics_scatter import ScatterGeodesicPlotter
-from einsteinpy.plotting.geodesics.geodesics_static import StaticGeodesicPlotter
+from einsteinpy.plotting.geodesics.geodesic_plotter import GeodesicPlotter
```
shreyasbapat: Files name would be 

core.py
static.py
interactive.py

src/einsteinpy/plotting/__init__.py
```diff
@@ -1,2 +1,6 @@
-from einsteinpy.plotting.geodesics.geodesics_scatter import ScatterGeodesicPlotter
-from einsteinpy.plotting.geodesics.geodesics_static import StaticGeodesicPlotter
+from einsteinpy.plotting.geodesics.geodesic_plotter import GeodesicPlotter
+from einsteinpy.plotting.geodesics.interactivegeodesic_plotter import (
+    InteractiveGeodesicPlotter,
+)
+from einsteinpy.plotting.geodesics.staticgeodesic_plotter import StaticGeodesicPlotter
+from einsteinpy.plotting.senile.geodesics_scatter import ScatterGeodesicPlotter
```
shreyasbapat: This line won't be here

src/einsteinpy/plotting/geodesics/geodesic_plotter.py
```diff
@@ -0,0 +1,22 @@
+import random
+
+from .interactivegeodesic_plotter import InteractiveGeodesicPlotter
```
shreyasbapat: No relative imports

shreyasbapat:
Where is the old staticgeodesic? That should also be put in senile

Varunvaruns9:
The old one is almost the same as the new StaticGeodesicPlotter. Do we need to keep that separately?

docs/source/api/plotting/geodesics/geodesics_scatter.rst
```diff
@@ -28,5 +28,5 @@ Color
 
 This module contains the methods for static geodesic plotting using scatter plots.
 
-.. automodule:: einsteinpy.plotting.geodesics.geodesics_scatter
+.. automodule:: einsteinpy.plotting.senile.geodesics_scatter
```
shreyasbapat: The path to the docs will also be changed!

shreyasbapat:
> The old one is almost the same as the new StaticGeodesicPlotter. Do we need to keep that separately?

Yes

src/einsteinpy/plotting/plotlyplotter.py
ritzvik: I would say that we should use `return_cartesian=True` when calling `calculate_trajectory`

OR

Use spherical to cartesian conversion using the `coordinates` module for `Schwarzschild`
And BL to cartesian for `Kerr`. To see the implementation just look at any one of the `calculate_trajectory` functions!

src/einsteinpy/plotting/geodesics/geodesic_plotter.py
Varunvaruns9: But if we import a plotting module from another plotting module that raises an error due to circular import.

Varunvaruns9: But if we try to import a plotting module from another plotting module it would give an error due to circular import. We have to give a relative path in either __init__.py or core.py.

docs/source/api/plotting/geodesics/geodesics_scatter.rst
shreyasbapat: I think you didn't change! The file should lie in 

docs/source/api/plotting/senile/geodesics_scatter.rst

Varunvaruns9:
@shreyasbapat Did most of the changes except the one with relative paths. Python gives an error due to circular import there.

shreyasbapat:
Okay!

PR #279, opened 2019-07-03
__str__ and __repr__ for Tensor Class

### LINKED ISSUE #250: __str__ and __repr__ for Tensor Class

OVSofia:
Fixes #250

### DIFF
--- a/src/einsteinpy/symbolic/tensor.py
+++ b/src/einsteinpy/symbolic/tensor.py
@@ -88,6 +88,7 @@ class Tensor:
         
 
         """
+
         if isinstance(arr, (list, tuple)):
             self.arr = sympy.Array(arr)
         elif isinstance(arr, sympy.Array):
@@ -120,6 +121,24 @@ class Tensor:
 
     def __getitem__(self, index):
         return self.arr[index]
+
+    def __str__(self):
+        """
+        Returns a String with a readable representation of the object of class Tensor
+
+        """
+        representation = "Tensor" + "\n"
+        representation += self.arr.__str__()
+        return representation
+
+    def __repr__(self):
+        """
+        Returns a String with a representation of the state of the object of class Tensor
+
+       """
+        interpretable_representation = self.__class__.__name__
+        interpretable_representation += self.arr.__repr__()
+        return interpretable_representation
 
     def tensor(self):
         """
--- a/src/einsteinpy/tests/test_symbolic/test_tensor.py
+++ b/src/einsteinpy/tests/test_symbolic/test_tensor.py
@@ -40,6 +40,23 @@ def test_Tensor_getitem():
         assert obj[p, q, r] - test_list[p][q][r] == 0
 
 
+def test_Tensor_str():
+    x, y, z = symbols("x y z")
+    test_list = [[[x, y], [y, x]], [[z, x], [y, z]]]
+    obj1 = Tensor(test_list)
+
+    assert "object at 0x" not in str(obj1)
+
+
+def test_Tensor_repr():
+    x, y, z = symbols("x y z")
+    test_list = [[[x, y], [y, sin(2 * z) - 2 * sin(z) * cos(z)]], [[z ** 2, x], [y, z]]]
+    obj1 = Tensor(test_list)
+
+    machine_representation = repr(obj1)
+    assert not "object at 0x" in machine_representation
+
+
 def test_TypeError1():
     arr = 0
     try:


src/einsteinpy/symbolic/tensor.py
```diff
@@ -31,8 +32,29 @@ def __init__(self, arr):
             raise TypeError("Only multi-dimensional list or Sympy Array is expected")
 
     def __getitem__(self, index):
+
         return self.arr[index]
 
+    def __str__(self) -> str:
```
ritzvik: Remove the `-> str`. It creates problem in some python versions.

src/einsteinpy/symbolic/tensor.py
```diff
@@ -31,8 +32,29 @@ def __init__(self, arr):
...
+        """
+        Returns a String with a readable representation of the object of class Tensor
+
+        """
+        str_with_the_result = "Tensor Object: "
+        for tensor in self.arr:
+            str_with_the_result += "\n  * {}".format(tensor)
+        return str_with_the_result
+
+    def __repr__(self) -> str:
```
ritzvik: Same here!

src/einsteinpy/symbolic/tensor.py
```diff
@@ -31,8 +32,29 @@ def __init__(self, arr):
             raise TypeError("Only multi-dimensional list or Sympy Array is expected")
 
     def __getitem__(self, index):
+
         return self.arr[index]
```
ritzvik: Why unnecessary newline here?

src/einsteinpy/symbolic/tensor.py
```diff
@@ -31,8 +32,29 @@ def __init__(self, arr):
...
 
+    def __str__(self) -> str:
+        """
+        Returns a String with a readable representation of the object of class Tensor
+
+        """
+        str_with_the_result = "Tensor Object: "
+        for tensor in self.arr:
+            str_with_the_result += "\n  * {}".format(tensor)
+        return str_with_the_result
```
ritzvik: I would say the best way to create `__str__` is use the `__str__` function of `self.arr` itself.
Something like:
```
s = "Tensor\n"
s += self.arr.__str__()
return s
```

src/einsteinpy/symbolic/tensor.py
```diff
@@ -31,8 +32,29 @@ def __init__(self, arr):
...
+
+    def __repr__(self) -> str:
+        """
+        Returns a String with a representation of the state of the object of class Tensor
+
+       """
+        interpretable_representation = "{self.__class__.__name__}({self.arr})".format(
+            self=self
+        )
+        return interpretable_representation
```
ritzvik: Same for `__repr__` as discussed in case of `__str__`

src/einsteinpy/tests/test_symbolic/test_tensor.py
```diff
@@ -40,6 +40,23 @@ def test_Tensor_getitem():
         assert obj[p, q, r] - test_list[p][q][r] == 0
 
 
+def test_Tensor_str():
+    x, y, z = symbols("x y z")
+    test_list = [[[x, y], [y, x]], [[z, x], [y, z]]]
+    obj1 = Tensor(test_list)
+
+    assert obj1.__str__ is not object.__str__
```
ritzvik: I did not understand what is implied by this test. Maybe reply to this comment explaining what you tested?

Or change this taking this as reference.
https://github.com/einsteinpy/einsteinpy/blob/master/src/einsteinpy/tests/test_coordinates/test_coord_transform.py#L104-L123

src/einsteinpy/tests/test_symbolic/test_tensor.py
```diff
@@ -40,6 +40,23 @@ def test_Tensor_getitem():
...
+
+
+def test_Tensor_repr():
+    x, y, z = symbols("x y z")
+    test_list = [[[x, y], [y, sin(2 * z) - 2 * sin(z) * cos(z)]], [[z ** 2, x], [y, z]]]
+    obj1 = Tensor(test_list)
+
```
ritzvik: Same here. Tell me your testing approach.

src/einsteinpy/symbolic/tensor.py
OVSofia: I was going to write a docstring, but then I thought that in that case it was not neccesary, but I've forgotten to delete the newline.

src/einsteinpy/tests/test_symbolic/test_tensor.py
OVSofia: my approach to test if t the __str__ method is implemented is to check if the method is different from the default. I test it in my local machine and it worked

src/einsteinpy/tests/test_symbolic/test_tensor.py
OVSofia: In my first approach I wanted to use the eval() function. But the codefactor gave error because it seems it is not a secure function. So as an alternative, I used the function the repr function which used the __repr__. Maybe I could improve this test by checking if the class name is in the string

OVSofia: like this 
`    machine_representation = repr(obj1)
    assert obj1.arr.__str__() in machine_representation`

OVSofia:
codecov solved!

src/einsteinpy/tests/test_symbolic/test_tensor.py
ritzvik: If we do `assert True` after a whole bunch of code, the tests would pass anyway. 
By default `object at 0x` appears when printing an object where `__str__` and `__repr__` is not defined. You have to make sure it does not appear.

src/einsteinpy/tests/test_symbolic/test_tensor.py
ritzvik: Kindly capture the output of `print(object_name)`, and do some tests on it!

src/einsteinpy/symbolic/tensor.py
```diff
@@ -33,6 +34,24 @@ def __init__(self, arr):
...
+        representation += self.arr.__str__()
+        return representation
+
+    def __repr__(self):
+        """
+        Returns a String with a representation of the state of the object of class Tensor
+
+       """
+        interpretable_representation = self.__class__.__name__
+        interpretable_representation += self.arr.__str__()
```
ritzvik: Call `self.arr.__repr__()` here.

Also it would be great if you could show some screenshots of the output.
I think, you didn't quite understand the utility of `__str__` and `__repr__`, I may be wrong.

src/einsteinpy/tests/test_symbolic/test_tensor.py
ritzvik: The tests should reflect the fact that it does not output the default thing `object at 0x.....blah blah` on doing `print(object)`

src/einsteinpy/tests/test_symbolic/test_tensor.py
OVSofia: image

As you can see in the screenshot, the output in the __str__ is human readable not like the output of object. What the test does it to check if the method __str__ of tensor class is not the default __str__ which I understand if it is not implemented, I would see something like "object x100..."

src/einsteinpy/tests/test_symbolic/test_tensor.py
OVSofia: image
The test reflects that there is not "object at 0x..." becuase I check if there is in the string the self.arr.__str__ is in the repr(tensor_object)

src/einsteinpy/symbolic/tensor.py
OVSofia: image
This is the result with __repr__
In fact I had  put __str__ here as in the other. something that I couldn't understood why. So now that you said me to change it, make sense again 

Correct me if I am wrong, but the method __str__ is used to print human readable information about the object and __repr__ if to give more technical information about it right?

src/einsteinpy/tests/test_symbolic/test_tensor.py
```diff
@@ -40,6 +40,23 @@ def test_Tensor_getitem():
...
+    machine_representation = repr(obj1)
+    assert obj1.arr.__str__() in machine_representation
```
ritzvik: `assert not "object at 0x" in machine_representation`

ritzvik: Because we don't want the default output

src/einsteinpy/tests/test_symbolic/test_tensor.py
ritzvik: `assert not "object at 0x" in str(obj1)`

src/einsteinpy/symbolic/tensor.py
```diff
@@ -121,6 +122,24 @@ def config(self):
     def __getitem__(self, index):
         return self.arr[index]
 
+    def __str__(self):
+        """
+        Returns a String with a readable representation of the object of class Tensor
+
+        """
+        representation = "Tensor" + "\n"
+        representation += self.arr.__str__()
```
cjayross: It should be noted that this does not support the pretty printing that sympy is designed for. Meaning that interactive environments will have a hard time understanding the values returned by these objects.

ritzvik: I know it would render as plaintext in the IPython notebook. Would keep it in mind. Meanwhile pretty printers can do `obj.tensor()`.

shreyasbapat:
Rebase! @OVSofia 

I believe it's good to go! Right @ritzvik

ritzvik:
Yups! Good to go.
I would say, rebase and merge @shreyasbapat .
It's one commit only!

PR #293, opened 2019-07-07
(Minor) Add an entry in .gitignore for Emacs

cjayross:
I use Spacemacs as my editor and one of the features that it implements is called projectile. This plugin allows easier navigation through source code by creating a file called TAGS which maps function names to file/line number so that the definitions for functions can be jumped to instantly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,9 @@ pip-wheel-metadata
 # VS Code
 .vscode
 
+# Emacs
+TAGS
+
 # default pytest cache directory
 */.pytest_cache
 .pytest_cache/


shreyasbapat:
Isn't this change too less for 2 commits? 
And this would need a rebase. 

And isn't this too specific to your case?

cjayross:
@shreyasbapat, that sounds likely, I'll see about rebasing.

And yes, it is. In my opinion though, it's a change that is as trivial as it's use case.

shreyasbapat:
No. Not really something that is needed. :)

Merging this

Issue #288, opened 2019-07-06
Extend support of Tensor Class(symbolic) to support scalars


ritzvik:
🐞 **Problem**
Scalars are also zero order tensors. The Tensor class currently supports 1st order tensors and above.

🎯 **Goal**

  - Make `Tensor` support scalars. This issue can also help in #287 .

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

OVSofia:
I'll handle this,  if that's all right

PR #301, opened 2019-07-19
EinsteinTensor

### LINKED ISSUE #300: Addition of Einstein Tensor to symbolic module
🎯 **Goal**
* Add a class EinsteinTensor in src/einsteinpy/symbolic/einstein.py
* https://en.wikipedia.org/wiki/Einstein_tensor

OVSofia:
Fixes #300

### DIFF
--- a/src/einsteinpy/symbolic/einstein.py
+++ b/src/einsteinpy/symbolic/einstein.py
@@ -1,5 +1,112 @@
-from .tensor import Tensor
+import sympy
+
+from einsteinpy.symbolic.ricci import RicciScalar, RicciTensor
+from einsteinpy.symbolic.tensor import Tensor, _change_config
 
 
 class EinsteinTensor(Tensor):
-    pass
+    """
+    Class for defining Einstein Tensor
+    """
+
+    def __init__(self, arr, syms, config="ll", parent_metric=None):
+        """
+        Constructor and Initializer
+
+        Parameters
+        ----------
+        arr : ~sympy.tensor.array.dense_ndim_array.ImmutableDenseNDimArray or list
+            Sympy Array or multi-dimensional list containing Sympy Expressions
+        syms : tuple or list
+            Tuple of crucial symbols denoting time-axis, 1st, 2nd, and 3rd axis (t,x1,x2,x3)
+        config : str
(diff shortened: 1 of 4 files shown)


src/einsteinpy/symbolic/einstein.py
```diff
@@ -1,5 +1,112 @@
...
+        """
+        Returns the Parent Metric, if available.
+        """
+        return self._parent_metric
+
+    @classmethod
+    def from_metric(cls, metric):
+        t_ricci = RicciTensor.from_metric(metric)
+        r_scalar = RicciScalar.from_riccitensor(t_ricci, t_ricci.parent_metric)
+        einstein_tensor = t_ricci.tensor() - 0.5 * metric.tensor() * r_scalar.tensor()
```
ritzvik: don't use `0.5`, use `(1/2)` instead. This is because, calculations are symbolic, so it will be better to have then in fractional form instead of decimal form.
Also, change `r_scalar.tensor()` to `r_scalar.expr`. Remember `RicciScalar` holds a single expression which can be obtained by using the property `expr`. It is not a tensor(It is actually a 0-order tensor, but we'll ignore the fact for now). For this, we would need to solve #288 xD

src/einsteinpy/symbolic/einstein.py
```diff
@@ -1,5 +1,112 @@
...
+        Exception
+            Raised when a parent metric could not be found.
+
+        """
+        if metric is None:
+            metric = self._parent_metric
+        if metric is None:
+            raise Exception("Parent Metric not found, can't do configuration change")
+        new_tensor = _change_config(self, metric, newconfig)
+        new_obj = RicciTensor(
```
ritzvik: **SILLY MISTAKE ALERT**
`new_obj = EinsteinTensor(....)`

OVSofia:
About the test, I was looking for testing the property of divergenless, but I am not sure on how calculate it. I have check and maybe CovarDerivativeOp could be usefull, but I'm not sure about how to use it :/ . Do you know another option for the test?

ritzvik:
This property can used to test https://en.wikipedia.org/wiki/Einstein_tensor#Trace

ritzvik:
@shreyasbapat 🔥 M.E.R.G.E.

PR #304, opened 2019-07-23
Geodesics: Add static 3D plotting.

### LINKED ISSUE #9: Add a 3D Plotting Module
Options: plotly, mayavi, mpl_3D

Varunvaruns9:
Fixes #9.

This adds 3D plotting for geodesics in matplotlib. I'll be adding the same for plotly soon.

Example images:

Isometric view:
Figure_1

Top View:
Figure_2

Front View:
Figure_3

@shreyasbapat @ritzvik Are these plots correct?

### DIFF
--- a/src/einsteinpy/plotting/geodesics/static.py
+++ b/src/einsteinpy/plotting/geodesics/static.py
@@ -4,10 +4,17 @@ import matplotlib.pyplot as plt
 import numpy as np
 from matplotlib.animation import FuncAnimation
 from matplotlib.patches import Circle
+from mpl_toolkits import mplot3d
 
 
 class StaticGeodesicPlotter:
-    def __init__(self, ax=None, attractor_radius_scale=-1.0, attractor_color="#ffcc00"):
+    def __init__(
+        self,
+        ax=None,
+        attractor_radius_scale=-1.0,
+        attractor_color="#ffcc00",
+        use_3d=False,
+    ):
         """
         Constructor.
 
@@ -23,21 +30,27 @@ class StaticGeodesicPlotter:
 
         """
         self.ax = ax
+        self.use_3d = use_3d
         if not self.ax:
             self.fig, self.ax = plt.subplots(figsize=(6, 6))
-            self.ax.set_xlabel("$x$ (km)")
-            self.ax.set_ylabel("$y$ (km)")
             self.ax.set_aspect(1)
+            if self.use_3d:
+                self.ax = plt.axes(projection="3d")
+                self.ax.set_zlabel("$z$ (m)")
+            self.ax.set_xlabel("$x$ (m)")
+            self.ax.set_ylabel("$y$ (m)")
         self.attractor_radius_scale = attractor_radius_scale
         self.attractor_color = attractor_color
         self.attractor_present = False
 
-    def _mindist(self, x, y):
-        return np.sqrt(x * x + y * y)
+    def _mindist(self, x, y, z=0):
+        return np.sqrt(x * x + y * y + z * z)
 
     def _draw_attractor(self, radius, xarr, yarr):
         self.attractor_present = True
-        if self.attractor_radius_scale == -1.0:
+        if self.use_3d:
+            self.ax.plot([0], [0], [0], "o", mew=0, color=self.attractor_color)
+        elif self.attractor_radius_scale == -1.0:
             minrad_nooverlap = self._mindist(xarr[0], yarr[0])
             for i, _ in enumerate(xarr):
                 minrad_nooverlap = min(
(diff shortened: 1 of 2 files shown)


ritzvik:
@Varunvaruns9 , the plots look cool, can you give the parameters for which the plot was made?

Varunvaruns9:
@ritzvik I took the values from the tests.
```
obj = SphericalDifferential(
    130 * u.m,
    np.pi / 2 * u.rad,
    -np.pi / 8 * u.rad,
    0 * u.m / u.s,
    0 * u.rad / u.s,
    1900 * u.rad / u.s,
)
att = Body(name="attractor", mass=6e24 * u.kg, parent=None)
b1 = Body(name="obj", differential=obj, parent=att)
t = 0 * u.s
start_lambda = 0.0
end_lambda = 0.002
step_size = 5e-8
geo = Geodesic(b1, time=t, end_lambda=end_lambda, step_size=step_size)
```

ritzvik:
@Varunvaruns9 , as the particle has a initial velocity in the equatorial plane only(see v_r and v_theta are zero), the body should remain in the same plane, and it is doing do (see that the z axis has units `1e-13`, so practically it is in the same plane). However, on a first glance, the plot looks terribly wrong due to autoscaling of matplotlib. Can something be about about it???

shreyasbapat:
What I would suggest is, it is right so let's include it anyways.

shreyasbapat:
Apart from that. @Varunvaruns9  have a look at `Figure` rather than a `FigureWidget` in Plotly 4.0 And tryto come up with a good plotly plotter

src/einsteinpy/plotting/geodesics/static.py
```diff
@@ -120,6 +125,31 @@ def _update(frame):
             self.fig, _update, frames=frames, interval=interval, blit=True
         )
 
+    def plot3d(self, geodesic, color="#{:06x}".format(random.randint(0, 0xFFFFFF))):
```
shreyasbapat: Instead of this, can we pass a parameter, `use_3d = True` or something like that in the __init__ ?

ritzvik:
> What I would suggest is, it is right so let's include it anyways.

We can keep a lower limit for scaling of each of the axis, say `1e-5`

Also, the distances are in `m`s, not `km`s as suggested by the markers.

shreyasbapat:
Plotly 4.0 suggests we are anyways not going to use matplotlib elsewhere xD

src/einsteinpy/plotting/geodesics/static.py
Varunvaruns9: Okay, I'll add both features in the same plot function.

shreyasbapat:
@Varunvaruns9 Let me know when you think you are done! I will do a thorough review.

Varunvaruns9:
@ritzvik I've added a limit for z-scaling, should I do this for x and y axes as well?

Varunvaruns9:
Also, can you help me with an example where there is significant movement in the z-direction as well? We have to set `v_r` and `v_t` to non-zero values. Right?

I tried some random values but the trajectory was always in a plane.

ritzvik:
> @ritzvik I've added a limit for z-scaling, should I do this for x and y axes as well?

Yups, it is can be the case that the particle always remains in x-z plane or y-z plane.
So, in my opinion, limit scaling to all the axis

ritzvik:
> Also, can you help me with an example where there is significant movement in the z-direction as well? We have to set `v_r` and `v_t` to non-zero values. Right?
> 
> I tried some random values but the trajectory was always in a plane.

In schwarzschild case, it is always expected to remain in a plane because, the schwarzschild geometry is spherically symmetric. You can always rotate the coordinates in such a way that `v_theta` can be 0, and all the velocity can be directed to `v_r` and `v_phi`.

shreyasbapat:
@Varunvaruns9 Any updates on this?

Varunvaruns9:
I added scaling for all axes but I had to use many if-else statements, so the codeclimate check is failing. Maybe we can increase the limit?

ritzvik:
> I added scaling for all axes but I had to use many if-else statements, so the codeclimate check is failing. Maybe we can increase the limit?

I think we can safely ignore codeclimate here!

src/einsteinpy/plotting/geodesics/static.py
```diff
@@ -60,6 +73,16 @@ def _draw_attractor(self, radius, xarr, yarr):
                 Circle((0, 0), radius.value, lw=0, color=self.attractor_color)
             )
 
+    def _set_scaling(self, x_range, y_range, z_range, lim):
+        if x_range < lim and y_range < lim and z_range < lim:
```
Varunvaruns9: I added this so that if all ranges are less than the limit then we should not change the scale of axes as everything would then become too small.

Varunvaruns9:
@shreyasbapat Please review.

src/einsteinpy/plotting/geodesics/static.py
shreyasbapat: RIght

Issue #309, opened 2019-07-28
Add loads of predefined metrics in symbolic module


ritzvik:
🐞 **What bugs you?**
  - Have predefined metrics in *symbolic* module

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

kauii8school:
I'm about to submit a pull request adding in the "Barriola Vilekin" metric. There are plenty of readily available metrics at https://arxiv.org/pdf/0904.4184.pdf. I will be adding those in the next few days!

ritzvik:
Reopening this.
This is gonna be a perpetual issue 😛

obi-wan-shinobi:
I added the Reissner-Nordstorm Metric. Please review so that I can add more :)

Issue #263, opened 2019-06-28
Integrate Sympy's Lambdify in EinsteinPy


shreyasbapat:
Our tensors should have the ability to represent arrays of numpy-like functions rather than sympy expressions

ritzvik:
@OVSofia , search about sympy's lambdify. Sooner or later, you would have to do this 😄

ritzvik:
@OVSofia , only a few days are left in the SOCIS timeframe, so can you start working on this?

OVSofia:
> @OVSofia , only a few days are left in the SOCIS timeframe, so can you start working on this?

I was with KerrMetric and KerrNewmanMetric, but I can fix this first

ritzvik:
> > @OVSofia , only a few days are left in the SOCIS timeframe, so can you start working on this?
> 
> I was with KerrMetric and KerrNewmanMetric, but I can fix this first

I have other plans for them. Let's work on this first!

OVSofia:
@ritzvik @shreyasbapat I think this task is already (in the absence of being reviewed ) done. I have seen that issue #259 is missing. Do I start with that?

Issue #259, opened 2019-06-27
Idea: Make Tensor(symbolic module) a derived class of sympy.Array


ritzvik:
🐞 **Problem**
As of now, we have to explicitly define already available functions in sympy like `subs()`, `__getitem__()` etc.

🎯 **Goal**
  - Making this a derived class of `sympy.Array` would save a lot of hassle. Idea open to suggestions.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

PR #311, opened 2019-07-28
Grand Base tensor for GR quantities

### LINKED ISSUE #310: Make another derived class from Tensor and make it base for all the tensors in symbolic module
🎯 **Goal**
  - Make another class `GRTensor` from `Tensor` and use it as a base class for `MetricTensor`, `ChristoffelSymbol` etc. This way we can transfer many of the repetitive functions like `change_config` into it.
Further, this would be a step towards #259

ritzvik:
Fixes #310

### DIFF
--- a/src/einsteinpy/symbolic/__init__.py
+++ b/src/einsteinpy/symbolic/__init__.py
@@ -4,6 +4,6 @@ from .metric import MetricTensor
 from .ricci import RicciScalar, RicciTensor
 from .riemann import RiemannCurvatureTensor
 from .stress_energy_momentum import StressEnergyMomentumTensor
-from .tensor import Tensor
+from .tensor import BaseRelativityTensor, Tensor
 from .vacuum_metrics import SchwarzschildMetric
 from .weyl import WeylTensor
--- a/src/einsteinpy/symbolic/christoffel.py
+++ b/src/einsteinpy/symbolic/christoffel.py
@@ -1,12 +1,13 @@
 import numpy as np
 import sympy
 
-from einsteinpy.symbolic.tensor import Tensor, _change_config
+from einsteinpy.symbolic.tensor import BaseRelativityTensor, _change_config
 
 
-class ChristoffelSymbols(Tensor):
+class ChristoffelSymbols(BaseRelativityTensor):
     """
-    Class for defining christoffel symbols
+    Inherits from ~einsteinpy.symbolic.tensor.BaseRelativityTensor .
+    Class for defining christoffel symbols.
     """
 
     def __init__(self, arr, syms, config="ull", parent_metric=None):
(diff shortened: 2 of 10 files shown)


ritzvik:
This PR lays the initial groundworkfor solving : 
  - #309 
  - #259 
  - #263 


Further, it reduces the code duplicacy, without breaking the API (for v0.2.1).

Some challenges : 
  - Could not implement `change_config` function as the base class `GRTensor` would not be able to return instances of inherited classes, when `change_config` is called from an instance of inherited class, say `ChristoffelSymbols`
  - Could not check(and raise error) that `parent_metric` argument is an instance of `MetricTensor` as it is a case of cyclic imports.

*These would need some workarounds, would open a few issues*

Otherwise, LGTM 
cc @shreyasbapat 

@OVSofia , have a thorough look at this PR. You might need to edit #308 after this, a little bit.

OVSofia:
> This PR lays the initial groundwork for solving :
> 
> * #309
> * #259
> * #263
> 
> Further, it reduces the code duplicacy, without breaking the API (for v0.2.1).
> 
> Some challenges :
> 
> * Could not implement `change_config` function as the base class `GRTensor` would not be able to return instances of inherited classes, when `change_config` is called from an instance of inherited class, say `ChristoffelSymbols`
> * Could not check(and raise error) that `parent_metric` argument is an instance of `MetricTensor` as it is a case of cyclic imports.
> 
> _These would need some workarounds, would open a few issues_
> 
> Otherwise, LGTM
> cc @shreyasbapat
> 
> @OVSofia , have a thorough look at this PR. You might need to edit #308 after this, a little bit.

Just minor changes yes. Anyway I will wait for this PRwas merge and my PR's correction to make those minor changes :)

src/einsteinpy/symbolic/__init__.py
```diff
@@ -4,6 +4,6 @@
 from .ricci import RicciScalar, RicciTensor
 from .riemann import RiemannCurvatureTensor
 from .stress_energy_momentum import StressEnergyMomentumTensor
-from .tensor import Tensor
+from .tensor import GRTensor, Tensor
```
shreyasbapat: Instead of GRTensor, name it BaseTensor :D 
That would be more appropriate!

src/einsteinpy/symbolic/christoffel.py
```diff
@@ -1,12 +1,13 @@
 import numpy as np
 import sympy
 
-from einsteinpy.symbolic.tensor import Tensor, _change_config
+from einsteinpy.symbolic.tensor import GRTensor, _change_config
 
 
-class ChristoffelSymbols(Tensor):
```
shreyasbapat: What will be the future of Tensor?

src/einsteinpy/symbolic/tensor.py
```diff
@@ -180,3 +181,123 @@ def simplify(self):
 
         """
         return sympy.simplify(self.tensor())
+
+
+class GRTensor(Tensor):
```
shreyasbapat: Okay, I change my words, rename it to `BaseRelativityTensor` or something like that!

src/einsteinpy/symbolic/christoffel.py
ritzvik: `Tensor` would refer to any tensor while this new one can solely be used to relativity purpose.
Don't know may deprecate `Tensor` with 0.3.0. Let's see

src/einsteinpy/symbolic/tensor.py
ritzvik: Correct! Doing it!

src/einsteinpy/symbolic/__init__.py
ritzvik: > A little consideration. Everything looks fine. Any hopes of getting green tick on this?'
Codeclimate can't be fixed. There are too many exceptions to throw !

ritzvik:
@shreyasbapat DONE

Issue #305, opened 2019-07-24
Adapt to Plotly 4.0


shreyasbapat:
🐞 **Problem**

There are a lot of new cool additions to plotly 4.0 
(https://medium.com/@plotlygraphs/plotly-py-4-0-is-here-offline-only-express-first-displayable-anywhere-fc444e5659ee)

🎯 **Goal**

Use `Figure` instead of widgets. Maybe that will show plots in docs, and we will release a minor version.

💡 **Possible solutions**

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Varunvaruns9:
The new version looks great. I'll start working on it after #304.

shreyasbapat:
> The new version looks great. I'll start working on it after #304.

Cool!

PR #318, opened 2019-08-04
Migrate to Plotly 4.0

Varunvaruns9:
Use Figure instead of FigureWidget.
Add require.js file to docs.

Fixes: #305.

@shreyasbapat Rebased to the lastest master.


shreyasbapat:
Good to go!

PR #327, opened 2019-08-21
[WIP] Speed up calculations.

Varunvaruns9:


Varunvaruns9:
I don't think we need this now :)

Issue #234, opened 2019-06-08
Add more Bodies to the bodies module


shreyasbapat:
🐞 **Problem**

Currently, the bodies module only has 2 bodies. The list can be extended.

🎯 **Goal**

A better plotting experience

💡 **Possible solutions**

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

abhijeetmanhas:
I will work on it. I have created classes for moon, 7 other planets of solar system and Pluto. I will be adding mass and radius details. I have already done for Moon.

SuyashSalampuria:
@shreyasbapat I would like to work on this issue .
Also, could you give details of what changes should I do to achieve this

PR #329, opened 2019-08-27
Newbodies added to bodies module

abhijeetmanhas:
Addresses #234 
I have added class for moon and other planets of solar system + Pluto to `bodies.py` .I have used all data from [NASA JPL](https://ssd.jpl.nasa.gov/?planet_phys_par).
@shreyasbapat @ritzvik please review it

### DIFF
--- a/src/einsteinpy/bodies.py
+++ b/src/einsteinpy/bodies.py
@@ -95,3 +95,15 @@ class _Earth(Body):
 
 
 Earth = _Earth()
+
+
+class _Moon(Body):
+    def __init__(self):
+        parent = Earth
+        name = "Moon"
+        R = 1737.5 * u.km
+        mass = 7.34767309e22 * u.kg
+        super(_Moon, self).__init__(name=name, mass=mass, R=R, parent=parent)
+
+
+Moon = _Moon()
--- a/src/einsteinpy/bodies.py
+++ b/src/einsteinpy/bodies.py
@@ -107,3 +107,99 @@ class _Moon(Body):
 
 
 Moon = _Moon()
+
+
+class _Mercury(Body):
+    def __init__(self):
+        parent = Sun
+        name = "Mercury"
+        R = 2439.4 * u.km
+        mass = 3.30114e23 * u.kg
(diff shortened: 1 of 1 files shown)


abhijeetmanhas:
Should I write tests for it too before merging?

rishi-s8:
@abhijeetmanhas indeed. Tests are important. :)

src/einsteinpy/bodies.py
```diff
@@ -95,3 +95,111 @@ def __init__(self):
...
+class _Neptune(Body):
+    def __init__(self):
+        parent = Sun
+        name = "Neptune"
+        R = 24622 * u.km
+        mass = 1.024126e26 * u.kg
+        super(_Neptune, self).__init__(name=name, mass=mass, R=R, parent=parent)
+
+
+Neptune = _Neptune
```
ritzvik: `Neptune = _Neptune()` ?

ritzvik:
@shreyasbapat 

I've not developed the bodies module, but still, here's my two cents.

  - It wasn't a wise design decision to have `BlahBlahBody = _BlahBlahBody()` because each time we import anything from `bodies` module, all 11 `__init__()` in all predefined bodies are getting executed, making the module slow! 
  - Further, we are not giving user the freedom to make different instances of the same class. We are giving them predefined instances. Now it may be possible that the user might want 2 different instances of `_Sun()`, one with a slight change in it, but he won't be able to do that easily.

shreyasbapat:
Well, it was never intended to have solar system bodies

abhijeetmanhas:
It was mentioned in comments of `bodies.py` that it was supposed to have predefined bodies, I added those. I also thought that writing `body = _body` will make it slow, but I found it was done for sun and earth, so I followed same format.
So should I remove all lines containing `body = _body` even for earth and sun? Or should I not create different classes for other planets, but store their attributes in some dictionary and  create 3 parent classes: star, planet & satellite?

src/einsteinpy/bodies.py
ritzvik: Can you make this change and afterwards, rebase the branch?
@abhijeetmanhas

PR #336, opened 2019-09-06
Add interactive 3d plotter

Varunvaruns9:
@shreyasbapat 

newplot

### DIFF
--- a/src/einsteinpy/plotting/geodesics/interactive.py
+++ b/src/einsteinpy/plotting/geodesics/interactive.py
@@ -7,7 +7,9 @@ from plotly.offline import plot as saveplot
 
 
 class InteractiveGeodesicPlotter:
-    def __init__(self, attractor_radius_scale=-1.0, attractor_color="#ffcc00"):
+    def __init__(
+        self, attractor_radius_scale=-1.0, attractor_color="#ffcc00", use_3d=False
+    ):
         """
         Constructor.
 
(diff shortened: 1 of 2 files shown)


Varunvaruns9:
codeclimate 😞

shreyasbapat:
Can we include scaling here also? 
And then some last basic steps such as putting appropriate units to the axes, labels and stuff.

Varunvaruns9:
There are a few problems in drawing the attractor in 3d. Since the scales of the axes are different, the sphere becomes a 3d ellipse. We can use a 3d ellipse with dimensions such that it looks like a sphere when drawn but neither Matplotlib nor Plotly has any inbuilt method to do that. We can draw it manually with a mesh but it would be costly to draw a smooth figure.

> And then some last basic steps such as putting appropriate units to the axes, labels and stuff.
Also, what else needs to be added? @shreyasbapat

shreyasbapat:
Right. Can you give me an example where it looks like that!

shreyasbapat:
Also, for some reasons not understandable to be there are merge conflicts

Varunvaruns9:
Figure_1
@shreyasbapat This is what we'll get if we try to draw a sphere.

shreyasbapat:
Can we make any progress with this?

JeS24:
#612 fixes the scaling / aspect issue.

PR #347, opened 2019-12-04
Raise error when coordinate conversion is impossible

### LINKED ISSUE #343: Couldn't not run Coordinate Conversion example / tutorial
🐞 **Problem**

Can't run Coordinate Conversion example with message RuntimeWarning: invalid value encountered in double_scalars in coordinates/utils.py:85

So when calling `pos_vec.to_bl` at python example beginning  
```
a = 0.5 * u.km
pos_vec = Cartesian(.265003774 * u.km, -153.000000e-03 * u.km,  0 * u.km) 
bl_pos = pos_vec.to_bl(a)
print(bl_pos)
...
```

it execute `cartesian_to_bl_novel(x, y, z, a)` in coordinates/utils.py
```
w = (x ** 2 + y ** 2 + z ** 2) - (a ** 2)
r = np.sqrt(0.5 * (w + np.sqrt((w ** 2) + (4 * (a ** 2) * (z ** 2)))))
theta = np.arccos(z / r)
...
```
First line seems resulted as
```
w = 265.003774 ** 2 + (-153.0) ** 2 + 0.0 ** 2 - 500.0 ** 2 = -156363.99976575692
```
With `z = 0` second line reduces to `r = np.sqrt(0.5 * (w + np.sqrt((w ** 2) )))`. Obviously with any negative `w` it result to `r = 0` which leads to zero division in `theta = np.arccos(z / r)`

Varunvaruns9:
Fixes #343 
@shreyasbapat Raised an error for the problem described in #343.

### DIFF
--- a/src/einsteinpy/coordinates/utils.py
+++ b/src/einsteinpy/coordinates/utils.py
@@ -58,6 +58,8 @@ def cartesian_to_bl(x, y, z, a, v_x, v_y, v_z):
     """
     w = (x ** 2 + y ** 2 + z ** 2) - (a ** 2)
     r = np.sqrt(0.5 * (w + np.sqrt((w ** 2) + (4 * (a ** 2) * (z ** 2)))))
+    if r == 0:
+        raise ValueError("Coordinate conversion not possible")
     theta = np.arccos(z / r)
     phi = np.arctan2(y, x)
     dw_dt = 2 * (x * v_x + y * v_y + z * v_z)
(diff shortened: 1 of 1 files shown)


shreyasbapat:
Alright. Can you answer my question on the other issue?

shreyasbapat:
Do you think this is needed anymore?

Varunvaruns9:
Do you mean after #351?

It fixes cartesian->spherical but we can use the same code for cartesian->bl and vice-versa. So, it can be closed.

The velocity conversion can still give a divide by zero error in some cases though.

shreyasbapat:
Velocity :/

shreyasbapat:
Imma head out

ritzvik:
What I suggest is, we should raise warnings instead of errors.

shreyasbapat:
> What I suggest is, we should raise warnings instead of errors.

Part of the reasons why it is not merged :p

shreyasbapat:
no more a case

Issue #360, opened 2020-01-05
Add Kerr and Kerr-Newman metric in Boyer-Lindquist coordinates to predefined metrics


ritzvik:
🐞 **Problem**
`symbolic/predefined/vacuum_solutions.py` needs Kerr and Kerr-Newman Metrics. Implement them similar to already implemented `Schwarzschild()` in the same file.

💡 **Possible solutions**

  - [x] Kerr (c737c8b)
  - [ ] Kerr Newman

The metrics can be found be simple search on wikipedia. Also implementation can be found in `utils/kerr_utils.py` and `utils/kerrnewman_utils.py`

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

sagnik-chatterjee:
Can I take up this issue ?

ritzvik:
Try this one.

Regards,
Ritwik

gimseng:
Hi I am interested in helping with this issue, but i am new to git and open source contribution. If its alright, could I take this issue? My background is in theoretical physics (GR), so I understand the symbolic/analytic side of the problem, but just wanted to learn and apply git and python codings. I am reading through the developer guide and the git intro at the moment, so I am hoping to deal with an actual issue ticket to apply my learning.

ritzvik:
Sure @gimseng ! Take up the issue whenever you are comfortable. Ask us when you are stuck. 

Also, kindly join our community Chatroom on Riot.
https://riot.im/app/#/room/%23einsteinpy:matrix.org

gimseng:
Thanks I will tackle this. If I understand it correctly, the Kerr one has been updated, so I just need to include the charge (Kerr-Newman) appropriately.

ritzvik:
Yups. That’s correct 

Regards,
Ritwik

gimseng:
Thanks! I am almost done now, but have two questions:

1. I need a new constant "eps0" (electric constant or permittivity of free space). I defaulted it to "constants.eps0", could somehow help me to find where in the codes should I update/add this values? 

2. I needed 'pi', do we do math.pi or its defined as constants.pi or something?

Sorry to ask trivial questions, since I am new to EinsteinPy, and I thought maybe I could use the opportunities to ask questions on this discussion area. Thanks !

[...]

PR #427, opened 2020-02-11
Added Kerr-Newman metric in Boyer-Lindquist coordinates to predefined metrics, issue #360

gimseng:
Added Kerr-Newman metric in Boyer-Lindquist coordinates to predefined metrics, issue #360
Fixes #360 

@ritzvik

PR #428, opened 2020-02-11
Added KerrNewman to from .vacuum_solutions

gimseng:
part of the fix for #360 issue


ritzvik:
Add it to #427 only :)

PR #432, opened 2020-02-20
Update Shadow.py

Relativist1:
Faster Integration Time
No warnings

@shreyasbapat @ritzvik

PR #433, opened 2020-02-20
Update shadow.py

Relativist1:
Fixes #432 >

### DIFF
--- a/src/einsteinpy/rays/shadow.py
+++ b/src/einsteinpy/rays/shadow.py
@@ -2,7 +2,7 @@ import warnings
 
 import astropy.units as u
 import numpy as np
-from scipy.integrate import quadrature
+from scipy.integrate import fixed_quad
 from scipy.interpolate import interp1d
 from scipy.optimize import newton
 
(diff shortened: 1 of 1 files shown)


Relativist1:
Can't believe it...
finally!

ritzvik:
@Relativist1 are you sure of this?
Because you seem to be moving from an adaptive method to a fixed step-size one.
Any particular reasons to do this?

ritzvik:
What I see is the `tol` tolerance parameter is too low in `quadrature` function by default. Maybe we can increase the tolerance, and get a corresponding increase in step-size, reducing the time taken.

shreyasbapat:
We need to move to a fixed step size method.

shreyasbapat:
This was toooooo slow

ritzvik:
Ok then, merge it :)

Regards,
Ritwik

Relativist1:
Actually no matter how much I increased the tolerance it wouldn't matter matter.
Fixed size was getting the samw results in no time. 
So thats why.!
Also it gives no warnings as it used to be and sometimes kernel died too!

ritzvik:
> Actually no matter how much I increased the tolerance it wouldn't matter matter.
> Fixed size was getting the samw results in no time.
> So thats why.!
> Also it gives no warnings as it used to be and sometimes kernel died too!

This sure is fast! I saw that

ritzvik:
Thanks Saurabh!

Issue #339, opened 2019-09-22
Relativistic Radiative Transfer in arbitrary spacetime


sbhkmr1999:
🎯 **Goal**

Solving the Relativistic Radiative Transfer first in Schwarzschild and Kerr spacetime and then for arbitrary spacetime, for Emission Intensity (Emissivity) , for an observer in equatorial plane  (for starters) and then at any arbitrary angle.

💡 **Possible solutions**

https://arxiv.org/abs/1801.10452
https://arxiv.org/abs/1602.03184
http://discovery.ucl.ac.uk/1426969/1/Younsi_Ziri_741337%20_Ziri_Thesis.pdf

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

sbhkmr1999:
Some theoretical literature on the relativistic radiative transfer equations:

http://articles.adsabs.harvard.edu//full/2008ChJAS...8..226W/0000226.000.html
https://arxiv.org/abs/1304.5691
https://arxiv.org/abs/1207.4234

ritzvik:
@Relativist1 Thanks for the issue. Looking forward to adding this to EPY.

Bhavam:
Hi all , so apparently going through the requirements of the radiative transfer equation turns out it requires a bunch of parameters in the spacetime objects to be implemented for eg: lorentz invariant emission and  absorption coefficients which i did not find in the related classes , can @Relativist1 or @ritzvik verify this problem so I can start working on something to help on this issue.

ritzvik:
Yups, you’re correct. No such classes are available as of now. Is you’re willing to go ahead with this “sort-of” mini-project, you would need to create the helper functions and classes along with it. Ofc, I’ll be there for assistance. Also, @Relativist1 , can you provide some related papers??

Regards,
Ritwik

Bhavam:
@ritzvik okay cool ill start with defining the required class methods and parameters for schwarzschild bodies initially and keep you posted with a PR.

Relativist1:
https://arxiv.org/abs/1207.4234
Here is what I found that can be done.

Bhavam:
[...]

PR #463, opened 2020-03-26
Radiative transfer 

Bhavam:
Contributions for issue #339 
Defined some basic structure to work on and added comments on what to do.

Basically can you review code and instruct any pre requisites i failed to add to an EPY contribution like namespaces etc. @ritzvik 

Also as of right now its a first draft please tell of any problems with code style , use of libraries as per EPY standard , for example I felt vector initializations follow an existing template from vector class so pls instruct on how to use that.
  
Acc to the paper the authors only constructed code for kerr metrics which use fluid state which also I fail to find in EPY , so is it practical to add that or just take it as input ?

For the radiative transfer eqn two main functionalities must be defined first being radiative integration and second to obtain emission coeffecients so for that we have to use RK45 but mapping the variables from the eqn to y parameter (fun param) is a bit confusing any help @Relativist1 .

Thanks for your time and help .


ritzvik:
Thanks for the PR @Bhavam , I'll give a good look tomorrow :)

Relativist1:
@Bhavam Can you elaborate more into that front what exactly are you trying to say by eqn to y parameter? Also referring to which paper?

Bhavam:
@Relativist1 this is the eqn i am talking about , from http://articles.adsabs.harvard.edu//full/2008ChJAS...8..226W/0000226.000.html
eqn
So basically solution for di/dlambda has to be obtained so I (intensity field) is to be mapped to parameter y of rk45 function's fun param and lambda as t param.
eq2
So what im confused to is solution to intensity field is supposed to be the answer or the differential is the answer that should be returned by the function radiative transfer.

src/einsteinpy/radiative_transfer.py
```diff
@@ -0,0 +1,56 @@
+import astropy.units as u
+import numpy as np
+
+from einsteinpy import constant
+from einsteinpy.coordinates import SphericalConversion
+from einsteinpy.integrators import RK45
+
+'''create a class photon which would include a position vector and a velocity vector'''
```
shreyasbapat: This is not needed

src/einsteinpy/radiative_transfer.py
```diff
@@ -0,0 +1,56 @@
...
+class radiative_transfer:
```
shreyasbapat: Class names should follow this: https://www.python.org/dev/peps/pep-0008/#class-names

src/einsteinpy/radiative_transfer.py
```diff
@@ -0,0 +1,56 @@
...
+    '''This class holds methods to compute pitch angle , emmission coeffecient and radiative integration '''
+    def __init__(self,photons,metric): 
+        self.photon_pos=photon.position_vector #list of postions of individual photons
```
shreyasbapat: Where is `photon` defined?

src/photon.py
```diff
@@ -0,0 +1,15 @@
+import astropy.units as u
+import numpy as np
+
+from einsteinpy import constant
+
+''' any ideas for any existing base class this can inherit ?'''
+
+class photon:
```
shreyasbapat: Do we really need this class?

Bhavam: I think a separate class would be more practical than defining the radiative transfer params in each of spacetime types so that any arbitrary spacetime would use this class to contain these params ?

src/einsteinpy/radiative_transfer.py
Bhavam: Right Ill be careful of that from now on.

src/einsteinpy/radiative_transfer.py
Bhavam: I am unsure of the coordinate system that might be needed , i thought spherical would be apt acc to the paper if you have any other suggestions ?

src/einsteinpy/radiative_transfer.py
Bhavam: As a separate class as of now, but I assume it can be put inside a related parent class i am unsure depending on whether we need this ?

Bhavam:
> I suggest we discuss this thoroughly before moving forward in the chat room. I have some issues with the code quality, and the design. Let's see how that goes.

Yes definitely , and also please point point out anything else that i need to work on as I said its basically a draft version so I can change as per the requirement.

Bhavam:
Hey sorry for the long absence can you @ritzvik and @shreyasbapat give me any further instructions , should we discuss this in riot chatroom ?
I have made workable versions of some functionalities but all the missing parameters are being force initialized as of now so I have some doubts regarding their placement which we might ascertain in the chatroom ?

PR #482, opened 2020-04-04
Fix typo in kerr.py

yauneyz:
Fixed a typo discovered while researching Null Geodesics project

If this is the kind of thing that doesn't deserve its own pull request just let me know


ritzvik:
@yauneyz thanks! Every little change goes through a PR

PR #491, opened 2020-04-12
Update 'Lambdify Symbolic Calculation' example in docs

allardbrain:
Addresses #490

@ritzvik requested for PR review

Part 1 of 2 for the issue has already been merged. This PR submits part 2 of 2, which is to update the corresponding example in the docs.


allardbrain:
Please confirm if `"ll"` is the right value. Thanks!

ritzvik:
That would be all.
Thanks @allardbrain , Merging .

PR #505, opened 2020-05-29
User Defined Metrics example

### LINKED ISSUE #239: Write and example notebook for dealing in matrices
Sometimes users need to define their own spacetime, we should somehow facilitate that by adding an example on how to do that!

vikranthpulamathi:
Fixes #239 

@shreyasbapat 
@ritzvik


shreyasbapat:
Is this symbolic?

vikranthpulamathi:
@shreyasbapat I did import the symbolic module for the various pre-defined metrics (einsteinpy.symbolic) . Could that be the issue? If yes, then what is the approach to be made?

shreyasbapat:
I am sorry for the delayed response. The issue was not for symbolic module :( 

The issue was intended for the use of metric module.

Issue #405, opened 2020-01-31
Improvements to symbolic module!


ritzvik:
🐞 **Problem**
Lots of features are currently not available in symbolic module. I'm listing them and would continue to expand this list as things strike me! When working on a specific addition, kindly open an issue referencing to this one, and then open a PR. Mainly, the modifications would be required in `BaseRelativityTensor` class.

🎯 **Goal**

  - [ ] Wedge Product
  - [ ] Exterior Derivative
  - [ ] Covariant Derivative
  - [ ] Derivative (although, plain derivative is not a tensor, just a (n+1) order array)
  - [ ] Native function for Tensor Product

abhijeetmanhas:
For covariant derivative, is it we have to consider levi-civita connection of reimannian geometry?

kauii8school:
This may be relavent: https://docs.sympy.org/latest/modules/diffgeom.html. Is it ok to just use Sympy methods or should we implement native methods for all of these?

ritzvik:
Yes ! This seems relevant. We would use the functions as much as possible. 
I didn’t know about this.

Regards,
Ritwik

Starninja666:
Can I work on this?

ritzvik:
> I have started working on this..!

This is a huge tranche of issues. Could you give some specificity about the issue you are solving?

bibek22:
Hey @ritzvik, should I reference this issue itself or create a new issue for each goals listed above ?

JeS24:
This may be a good reference: http://gravity.psu.edu/local/internal/blackholes_2012.pdf (Check the appendix and first few chapters).

PR #513, opened 2020-06-12
Sym/derivative

bibek22:
Addresses `Derivative` item on  #405 
@shreyasbapat 

I tried to be flexible with the argument types. What do you think?  `_parse_vector()` was separated because that portion will be reused in implementing covariant derivative. I haven't added tests for all possible errors raised (there are too many), nor listed them all in the docstring. Let me know if I should.


shreyasbapat:
For me to start with a review, you will first need to pass the quality checks

bibek22:
> For me to start with a review, you will first need to pass the quality checks


Hey, most of the checks are passing now. What do I do about the codeclimate?

shreyasbapat:
> > For me to start with a review, you will first need to pass the quality checks
> 
> Hey, most of the checks are passing now. What do I do about the codeclimate?

The cognitive and cyclomatic complexities are very high in the function `_parse_vector`. Although, I am a bit skeptical about whether it should be in the place or not. But something has to be done about it.

I will review in a while.

shreyasbapat:
@OVSofia Can you please review this PR?

OVSofia:
> @OVSofia Can you please review this PR?

Sure, I will take a look at it tomorrow

src/einsteinpy/symbolic/tensor.py
```diff
@@ -116,6 +116,29 @@ def tensor_product(tensor1, tensor2, i=None, j=None):
     )
 
 
+def _parse_vector(x, syms):
+    # make a proper vector (tensor of config 'u') out of given x
+    if isinstance(x, BaseRelativityTensor):
+        if x.config == "u" and x.syms == syms:
+            return x
+        raise ValueError("Expected a tensor of config 'u' with syms %s" % syms)
+
```
OVSofia, line 125: As the test suggests, the problem is about the number of execution paths through code. So, maybe we should keep this function simple and directly do not enter in `_parse_vector` if `x` is already a `BaseRelativityTensor`,` x.config == "u" ` ... , what I suggest is make this check in `def Derivative()` (line 504). 

So I will delete this from `_parse_vector` :
```
if isinstance(x, BaseRelativityTensor):
        if x.config == "u" and x.syms == syms:
            return x
        raise ValueError("Expected a tensor of config 'u' with syms %s" % syms)
```
Let me know if it works!

OVSofia:
Cognitive Complexity is already solved but to make Cyclomatic complexity lower you could break the function up.  The more paths of execution the code has, the harder it can be to understand, so it is needed to try to avoid some 'ifs'.
Btw, I will take a look about the 'ifs' later, to try to give you a more precise solution!

src/einsteinpy/symbolic/tensor.py
```diff
@@ -501,7 +496,13 @@ def derivative(self, x):
         TypeError
             If argument ``x`` is of type other than listed above
         """
-        x = _parse_vector(x, self.syms)
+        if isinstance(x, BaseRelativityTensor):
+            if x.config != "u" and x.syms != self.syms:
+                raise ValueError(
```
OVSofia, line 501: Try to do the conditional block like this:
```
        if isinstance(x, BaseRelativityTensor) and x.config != "u" and x.syms != self.syms:
            raise ValueError( ...
```

src/einsteinpy/symbolic/tensor.py
```diff
@@ -118,11 +118,6 @@ def tensor_product(tensor1, tensor2, i=None, j=None):
...
 def _parse_vector(x, syms):
     # make a proper vector (tensor of config 'u') out of given x
-    if isinstance(x, BaseRelativityTensor):
-        if x.config == "u" and x.syms == syms:
-            return x
-        raise ValueError("Expected a tensor of config 'u' with syms %s" % syms)
-
     if isinstance(x, list):
         arr = x
         if len(x) != len(syms):
```
OVSofia, line 123: Let's keep the class type check separated from the length check of the list:

```
if isinstance(x, list):
        arr = x
elif isinstance(x, sympy.Symbol):
        arr = [1 if i == x else 0 for i in syms]
else:
         raise TypeError("Unexpected argument x of type %s" % type(x)) 

if len(arr) != len(syms):
         raise ValueError(
                "Expected length of vector %s, got %s" % (len(syms), len(x))
            )
x = BaseRelativityTensor(arr, syms=syms, config="u")

```

src/einsteinpy/symbolic/q
```diff
@@ -0,0 +1,458 @@
+import numpy as np
```
OVSofia, line 1: And why all these new files?

bibek22: whoops! looks like I accidentally git added unwanted files. I'll remove them. sorry.

src/einsteinpy/symbolic/tensor.py
bibek22: Looks like it's adding +1 to `and`/`or` operators too. If this is really important, I suggest to drop support for list type argument. It's one line of code on users part to convert it to `BaseRelativityTensor` anyway. What else could we do besides split it up into more functions?

tests/test_symbolic/test_tensor.py
```diff
@@ -281,3 +281,41 @@ def test_lambdify_with_args():
...
+    test_list = [[1, 0], [0, 1]]
+    obj2 = BaseRelativityTensor(test_list, syms=[x, y], config="ll")
+    # call using symbol
+    derived = obj1.derivative(x)
+    assert derived.arr == obj2.arr
+    # call using list
+    derived = obj1.derivative(x)
+    assert derived.arr == obj2.arr
+    # call using BaseRelativityTensor
+    test_list = [1, 0]
```
OVSofia, line 300: I think you try to test something else, but you forgot it and repeted the code:

`    # call using list
    derived = obj1.derivative(x)
    assert derived.arr == obj2.arr`

src/einsteinpy/symbolic/tensor.py
```diff
@@ -453,3 +469,52 @@ def lorentz_transform(self, transformation_matrix):
...
+            raise ValueError("Expected a tensor of config 'u' with syms %s" % self.syms)
+        if not isinstance(x, BaseRelativityTensor):
+            x = _parse_vector(x, self.syms)
+
+        arr = np.zeros(self.arr.shape, dtype=int)
+        arr = sympy.Array(arr)
+        for i in range(len(self.syms)):
+            if x[i]:
+                arr += x[i] * self.arr.diff(self.syms[i])
+
```
OVSofia, line 511: - I think this code would fail if you pass as an argument a list of one symbol, e.g :
```
obj1.derivative([x])

```
Maybe you should test it in order to check if fails.

- Moreover, this code will work one variable derivative, but have you contemplate the option of operating with multivariable-derivative (e.g.  ∂/(∂x ∂y))?

shreyasbapat:
Okay, please bear with me. I am very busy for next few days. @bibek22 I suggest you to attend our next community meeting. It will be on Friday 19:00 IST (Please adjust the timezone). Let's review after that.

src/einsteinpy/symbolic/tensor.py
bibek22: >     * I think this code would fail if you pass as an argument a list of one symbol, e.g :
> 
> 
> ```
> obj1.derivative([x])
> ```

This is true. For one symbol argument, one could simply pass in the symbol. 
`obj1.derivative(x)`

>    (e.g.  ∂/(∂x ∂y))?

By, this did you mean second partial derivative? If so, you'd call `.derivative()` twice:
`obj1.derivative(y).derivative(x)`

If you mean like directional derivative along certain direction other than coordinate axes, that's when you'd want to pass a `BaseRelativityTensor`.  Isn't that right ?

tests/test_symbolic/test_tensor.py
bibek22: yep, totally. I'll fix this in the next commit. Thanks!

src/einsteinpy/symbolic/tensor.py
OVSofia: Yes, by doing '.derivative' twice works. And this can be also a possible test to indicate how to make second partial derivative (instead of .derivate(x,y) or similar)

OVSofia: And sure "For one symbol argument, one could simply pass in the symbol", but it is better to know when it can fail.

bibek22: > but it is better to know when it can fail.

Right. So, maybe I should add this as a test case.

shreyasbapat:
Sorry for the long delay. Can you please rebase once?

src/einsteinpy/symbolic/tensor.py
shreyasbapat: Satisfactory

src/einsteinpy/symbolic/tensor.py
```diff
@@ -116,6 +116,22 @@ def tensor_product(tensor1, tensor2, i=None, j=None):
     )
 
 
+def _parse_vector(x, syms):
```
shreyasbapat, line 119: Please add a docstring to it

PR #512, opened 2020-06-11
Refactor - Metric & Utils

JeS24:
Fixes #113, #141, #410, #507, #508, #514 , #515.
Notes on fixes:
1. #508 - Requires discussion
2. #507 -`time_velocity` methods, across `<Metric>_utils` modules, have been [moved to and repackaged into 1 function in `coordinates.utils`](https://github.com/JeS24/einsteinpy/blob/0f90cb36d9f90e51659f3ddf204e6cf0302121c4/src/einsteinpy/coordinates/utils.py#L209-L250), renamed as `v_t()`. This function uses the supplied metric and 3-Velocity, and the expression for spacetime interval: $g_{ab} u^a u ^b = 0 or 1$ (null-like or time-like), to determine the timelike component of 4-Velocity, and hence it bypasses the method used in previous `Schwarzschild`, `Kerr` and `KerrNewman` classes.

#### Touched modules:
* `einsteinpy.metric`
* `einsteinpy.utils`
* `einsteinpy.coordinates.utils`
*  `geodesic`
* `examples`
* `plotting.geodesic`
* `test_metric`
* `test_utils`
* `test_coordinates/test_utils`
* `test_geodesic`
* `test_examples`
* `test_plotting/test_geodesic`

##### Other files
* `index.ipynb`
* `docs/source/jupyter.rst`
* `docs/source/user_guide`
* `docs/source/api/`
* `docs/source/examples/`
* `.codeclimate.yml`
* `CHANGELOG`

#### Current changes:
1. New `Metric` class added and old `metric` API removed.
2. Moved `calculate_trajectory()` from `metric` to `geodesic` and `time_velocity()` (now `v_t()`) to `coordinates.utils`.
3. Merged most utility functions from `utils` into `metric` (only file untouched is `scalar_factor.py`)
4. Modified `geodesic`, `examples` and example Jupyter Notebooks to work with new API.
5. Added some new utilities to `coordinates.utils`.
6. Added tests for new API (`metric`, `utils`, `geodesic`, `examples`, `plotting`)
7. Moved `calculate_trajectory()` tests from `test_metric` to `test_geodesic`.

#### Pending changes:
- [x] Alter code, after discussion on #508 - Factor removed.
- [x] Modify examples to use new API
- [x] Add functionality of `calculate_trajectory` to `geodesic` and create a warning in metric classes.
- [x] Add tests (can refactor old tests).
- [x] Fix CodeFactor and CodeClimate issues. - Linked to #113.
- [x] Add a changelog.
[...]

PR #521, opened 2020-06-26
Refactor coordinates and add support for 4-Vectors

### LINKED ISSUE #517: einsteinpy.bodies - Spin Factor, a is redundant
🐞 **Problem**
The `Body` class in `einsteinpy.bodies` uses a Spin Factor, `a`, as a parameter for the body. There are two issues with this:
1. `a` should be dimensionless. It's `alpha` or the rotational length parameter, that has the dimensions of length, and not `a`. The expressions for these two are: `a = Jc / GM**2` and `alpha = J / Mc = GMa / c**2`. `a` is used to define a constraint on the angular momentum, `J`. Also see Issue #514. 
2. More importantly, both of these parameters are relevant while defining a metric. If we are operating in the purview of vacuum solutions, an external body's mass, spin or charge is irrelevant, as they are incorporated in the metric itself. So, we can perhaps remove `a` from the `Body` class. Mass and charge can remain, as the purpose of the module seems to be to define classical bodies under classical limits, where constraining angular momentum is not a concern.

🎯 **Goal**

To make the purpose of the module clearer.

💡 **Possible solutions**

Discuss this and then decide on the removal of the factor.

 * Submit a pull request.
 * Add this in the comments:
 	- `Addresses #<put issue number here>` if you are partially fixing the issue.
 	- `Fixes #<put issue number here>` if you are completely fixing the issue.

### LINKED ISSUE #523: Math issues across metric modules in einsteinpy.metric
🐞 **Problem**

All terms of the metrics in the metric module are divided by `1 / c**2`. This is incorrect. I am linking the lines of code here, along with a Wiki page to compare them to:
1. Schwarzschild - [`metric.Schwarzschild`](https://github.com/einsteinpy/einsteinpy/blob/78bf5d942cbb12393852f8e4d7a8426f1ffe6f23/src/einsteinpy/metric/schwarzschild.py#L55-L59) - [Wikipedia](https://en.wikipedia.org/wiki/Schwarzschild_metric#The_Schwarzschild_metric)
2. Kerr - [`metric.Kerr`](https://github.com/einsteinpy/einsteinpy/blob/78bf5d942cbb12393852f8e4d7a8426f1ffe6f23/src/einsteinpy/metric/kerr.py#L126-L140) - [Wikipedia](https://en.wikipedia.org/wiki/Kerr_metric#Boyer%E2%80%93Lindquist_coordinates)
3. Kerr-Newman - [`metric.KerrNewman`](https://github.com/einsteinpy/einsteinpy/blob/78bf5d942cbb12393852f8e4d7a8426f1ffe6f23/src/einsteinpy/metric/kerrnewman.py#L135-L145) - [Wikipedia](https://en.wikipedia.org/wiki/Kerr%E2%80%93Newman_metric#Boyer-Lindquist_coordinates)

Metric, `g`, is defined as `ds^2 = c^2 * d\tau^2 = g_{ab}dx^adx^b`. But in the code, `ds^2` seems to have been equated with `d\tau^2`, but that cannot be the case, until `c` is set to 1.

The metric derivative (`_dg_dx`) and Christoffel symbol (`_ch_sym`) methods also suffer from the same issue.

🎯 **Goal**

To fix mathematical inaccuracies in the package.

💡 **Possible solutions**

Check the metric definitions and make corresponding changes. The issue has been confirmed after a review. The fix will be attempted in two steps:
- [x] Fix the metric definitions (`g_cov` or `metric_covariant`), so that the calculation of the time-like component of 4-Velocity is correct.
- [ ] Fix metric derivatives (`_dg_dx`) and Christoffel symbol (`_ch_sym`) methods, so that overall calculation is correct.

JeS24:
Fixes #517 

Things to do:
- [x] Write tests for refactor
- [x] Update `examples`, `metric`, Example Notebooks (wherever `coordinates` were removed in #512).
- [ ] Update `CHANGELOG`

### Mini-Changelog:
* `einsteinpy.coordinates`:
    - Renamed `coordinates.velocity` to `coordinates.differential`
    - Added 4-Vector support to `core` and `differential` (formerly, `velocity`)
    - Added support for calculation of time-like component of 4-Velocity
<br>

* `einsteinpy.metric`:
    - Added support for `einsteinpy.coordinates.differential`
    - Added support for `astropy.units`
    - Refactored `metric.Schwarzschild` and `metric.KerrNewman`
    - Partially fixed #523:

Update 1: **`Schwarzschild` is error-free now. Metric definitions in `Kerr` and `KerrNewman` are error-free too. Remaining methods (`_christoffels`, `_dg_dx`, `f_vec`) are yet to be altered.**

Update 2: **`Schwarzschild` and `Kerr` are error-free now.  `KerrNewman` is yet to be altered.**

Update 3: **`KerrNewman`'s `f_vec` needs to be cross-checked. Remaining methods have been corrected.**

<br>

* `einsteinpy.geodesic`:
    - Added support for `einsteinpy.coordinates.differential`
    - Added support for `astropy.units`
    - Moved old code to `Geodesic` class, while Time-like geodesic calculation is now through `TimelikeGeodesic` class (which derives from `Geodesic`)
    - Added `state` class variable, that stores the initial state of the particle.
<br>

* `einsteinpy.units`:
    - Added a `primitive()` function, that helps in obtaining float values out of `astropy.units.quantity.Quantity` objects.
    - Removed `astro_dist` & `astro_sec` (These are just scaling factors and these are not being used anywhere at the moment - Removed to get better codecov diff score)
<br>

* `einsteinpy.bodies`:
    - Refactored bodies to remove `a` (Fixes #517) and some minor changes.
<br>

* Fixed #517, #524, partially fixed #523 
* Added new tests and refactored old tests, wherever applicable
* Refactored example notebooks


src/einsteinpy/bodies.py
```diff
@@ -31,62 +31,49 @@ def __init__(
...
             Radius of the body
-        differential : ~einsteinpy.coordinates, optional
-            Complete coordinates of the body
-        a : ~astropy.units.m, optional
-            Spin factor of massive body. Should be less than half of schwarzschild radius.
-        q : ~astropy.units.C, optional
-            Charge on the massive body
-        is_attractor : Bool, optional
-            To denote is this body is acting as attractor or not
+        state_vec : ~einsteinpy.coordinates, optional
```
shreyasbapat, line 50:
```suggestion
        state : ~einsteinpy.coordinates, optional

```

JeS24: This is done.

JeS24:
@shreyasbapat Is CodeClimate supposed to check `tests/`?

JeS24:
MyPy needs to be less aggressive with type checking.

JeS24:
The tests would fail most probably. @shreyasbapat There were the issues, I was talking about. However, I am attempting a minor fix/refactor, which may fix these, without removing `astropy.units`.

JeS24:
@shreyasbapat PR is ready for review now.

shreyasbapat:
I will review in a while. I am very busy for a couple of days. You can start working on other things for now.

JeS24:
@shreyasbapat Yeah, I am currently working on Null Geodesics, working out the equations and all.

src/einsteinpy/coordinates/conversion.py
```diff
@@ -1,127 +1,450 @@
...
+        Converts to Spherical Polar Coordinates
+
+        Other Parameters
+        ----------------
+        **kwargs : dict
+            Keyword Arguments
+
+        Returns
+        -------
+        ~numpy.ndarray
```
JeS24, line 85: Should be `tuple`. Fixed.

docs/source/api/coordinates/differential.rst
```diff
@@ -1,8 +1,8 @@
-velocity module
-===============
+differential module
```
shreyasbapat, line 1: Okay  makes sense

src/einsteinpy/bodies.py
```diff
@@ -15,10 +15,9 @@
 Data references can be found in :py:mod:`~einsteinpy.constant`
 """
 
-import astropy.units as u
+from astropy import units as u
```
shreyasbapat, line 18: Was this change really needed?

src/einsteinpy/examples.py
```diff
@@ -1,13 +1,14 @@
 import numpy as np
+from astropy import units as u
 
-from einsteinpy.coordinates.utils import four_position, stacked_vec
-from einsteinpy.geodesic import Geodesic
+from einsteinpy.coordinates import SphericalDifferential
+from einsteinpy.geodesic import TimelikeGeodesic
```
shreyasbapat, line 5: Okay this will change , but not now

src/einsteinpy/geodesic.py
```diff
@@ -34,29 +42,84 @@ def __init__(
...
 
     @property
     def trajectory(self):
+        """
+        Returns the "Trajectory" of the Geodesic
+
+        """
         return self._trajectory
 
+    def _calculate_state(self):
```
shreyasbapat, line 94: Is this being used?

src/einsteinpy/geodesic.py
```diff
@@ -157,28 +220,28 @@ def calculate_trajectory_iterator(
...
 
-            else:
-                conv_coords = {
-                    "S": SphericalConversion,
-                    "BL": BoyerLindquistConversion,
-                }
+        while True:
+            if return_cartesian:
+                # Getting the corresponding Conversion class
+                type_coords_conv = type(self.coords).__mro__[1]
```
shreyasbapat, line 235: Not a very clean code

src/einsteinpy/bodies.py
JeS24: That's due to `isort` changes.

src/einsteinpy/geodesic.py
JeS24: Based on the conversation in the chat room:
```
coords = self.coords
# Proxy object of the Conversion superclass
conv = super(type(coords), coords)
cart_vecs = list()
for v in vecs:
    conv.__init__(
        v[0], v[1], v[2], v[3], v[5], v[6], v[7]
    )
    vals = conv.convert_cartesian(M=g.M, a=g.a)
    cart_vecs.append(np.hstack((vals[:4], v[4], vals[4:])))
```
This method is object-oriented and coordinate-agnostic. And just a tad slower than `mro` or `bases`.

src/einsteinpy/examples.py
JeS24: I have changed the name to `Timelike`, as it's pretty evident, that we are talking about a time-like geodesic, when we write: `from einsteinpy.geodesic import Timelike`. As for Null Geodesics, we can either go with `Nulllike` or `NullGeodesic`, although the former is more "symmetrical" to `Timelike`, so to speak.

src/einsteinpy/geodesic.py
shreyasbapat: I feel its doable. We just need to put one/two extra lines of comment explaining the process.

JeS24: Moving back to previous code, using `bases`. Both have similar performance.

Issue #367, opened 2020-01-10
Frame Dragging Example breaks on New plotting code


shreyasbapat:
🐞 **Problem**

Fix the scaling on y axis :/ 

![newplot (1)](https://user-images.githubusercontent.com/24257914/72183374-de7d5480-3413-11ea-87eb-4f36201a725b.png)

See the issue. The plot shoild have been like this: 

![download](https://user-images.githubusercontent.com/24257914/72183403-efc66100-3413-11ea-9e1a-a813287c3359.png)

🎯 **Goal**

💡 **Possible solutions**

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

ritzvik:
It would break. See the scaling on the old and new y-axis.

shreyasbapat:
Oh :/ Didn't pay attention there lol

ritzvik:
Close this ? @shreyasbapat

ritzvik:
> Close this ? @shreyasbapat

You edited the description, I didn't see. Maybe edit the issue title also?

shreyasbapat:
No, the title is consistent. See this: 
![Screenshot from 2020-01-15 17-56-42](https://user-images.githubusercontent.com/24257914/72433955-7edeca80-37c0-11ea-9e54-46a7147108ef.png)

I am investigating the scaling in both the plotting codes.
This is a big issue since, we are not able to deal with interactive plots so well.

ritzvik:
We would need atleast the matplotlib code to work correctly to include the plots in EPY paper. Can we raise it to urgent status, @shreyasbapat ?
Also this needs to sorted before `0.2.2`

https://gist.github.com/ritzvik/9271334525e5e1044ceb9bba7b88143c

shreyasbapat:
I am really on it. Lot of things to do. Will get it sorted asap

_______________________________________
This email is sent from a smartphone. Please ignore grammatical mistakes
and typographical errors.
_______________________________________

ritzvik:
I finally pushed myself to dive into the geodesic plotting codebase, and I would start off by saying that the way, `GeodesicPlotter` does switching is ingenious.

Now coming to the issue, scaling is set to one in both the static and interactive 2D plotter, which does not suffice for this case.

[...]

Issue #122, opened 2019-03-17
Benchmark and Validate the core Schwarzschild and Kerr Algorithms


shreyasbapat:
🐞 **Problem**

Perform benchmark testing for both the algorithms. Because how do anyone know that the algorithms are correct? 

🎯 **Goal**

To get clear understanding on performance of algorithm and its validation.

💡 **Possible solutions**

Validate against the industry standard or the books etc. Publish clear cut benchmarking results.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

JeS24:
I am currently validating the methods in `einsteinpy.metric.Kerr` and `einsteinpy.metric.Schwarzschild`, as this is necessary to ensure the correctness of the null geodesics. We can add whatever tests result from this to `test_metric`.

JeS24:
@shreyasbapat  Does "benchmark" refer to how close our results are to published studies? In that case, the calculation and plots, that will be obtained from the GSoC additions to `geodesics` and `rays`, can perhaps serve as a benchmark.

shreyasbapat:
> @shreyasbapat Does "benchmark" refer to how close our results are to published studies? In that case, the calculation and plots, that will be obtained from the GSoC additions to `geodesics` and `rays`, can perhaps serve as a benchmark.

Yes. Precisely

Issue #105, opened 2019-03-11
Null Geodesics for Kerr & Schwarzschild Geometry


ritzvik:
🐞 **What bugs you?**
Null-geodesics is the path traced by light rays in 4-D space-time. We aim to calculate null geodesics around non-rotating (Schwarzschild) and rotating (Kerr) massive bodies given the initial conditions.
This feature would support the one of the celebrated phenomenons in GR, gravitational lensing.

🎯 **Goal**
  - Null geodesic in Kerr/Schwarzschild space-time
  - Plotting class for both Kerr and Schwarzschild null geodesics.

💡 **Possible solutions**
Papers for references :
  - https://arxiv.org/abs/1601.02063
  - https://www.aanda.org/articles/aa/abs/2004/36/aa0814/aa0814.html
  - http://www.math.mcgill.ca/gantumur/math599w17/project-kerr.pdf

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

ritzvik:
Some freely available repos related to the issue :
  - https://github.com/grtrace/grtrace
  - https://github.com/Varcho/GeneralRelativity

ritzvik:
Units `c=1,G=1` handling table https://en.wikipedia.org/wiki/Geometrized_unit_system

ritzvik:
Don't know if this would help https://github.com/martinberoiz/BHBattle, but it says it uses Kerr Null-geodesic generator.

ritzvik:
Related repo : https://github.com/rantonels/starless

ritzvik:
A related paper : https://iopscience.iop.org/article/10.1088/0067-0049/207/1/6

ritzvik:
https://iopscience.iop.org/article/10.1088/0004-637X/696/2/1616

shilpi1958:
YNOGK
Paper: https://arxiv.org/pdf/1305.1250.pdf
Code : http://www1.ynao.ac.cn/~yangxl/yxl.html

VedantParanjape:
Hey, i am coming from gsoc page, I'm am from electronics engineering background, with decent coding skills, does this problem idea require extensive physics knowledge?

ritzvik:
[...]

Issue #40, opened 2019-02-07
Add a 7th or 8th order ODE Solver


shreyasbapat:
Resources are available on internet. Also, they can be shared later on this thread.

sagnik-chatterjee:
Can i try it with a scipy , with using function DOP853 ..a eightth order runge-kunta ode solver ?

ritzvik:
I guess you should try out #360 first, then dive into deeper, not so good first issues. If you wanna go for it, be my guest :)

Regards,
Ritwik

shreyasbapat:
Truly, I agree with @ritzvik .
If you feel you know things about ODEs. You are welcome to take this up. Otherwise it's not a good first issue

SuyashSalampuria:
@shreyasbapat This seems interesting to me
Shall I try this one

shreyasbapat:
Sure, let's first discuss the whole problem somewhere. And then you can proceed!

JeS24:
Some related literature:
* https://arxiv.org/pdf/2010.02237.pdf
* https://www.unige.ch/~hairer/poly_geoint/week2.pdf

PR #548, opened 2021-01-03
Improving Geodesic Integration in Python

### LINKED ISSUE #535: Improve the python back-end in `geodesic`
🐞 **Problem**

Currently, the python back-end in `geodesic` is not accurate for long simulations, unlike the `julia` back-end. Case in point, consider the following plots for a Kerr "Closed" geodesic, plotted using python and julia back-ends respectively:

pyth-bad
Closed Orbit Julia SGPL

Clearly. the geodesic, produced by the python back-end is not accurate. This is a problem, that plagues all "long" simulations, done using the python back-end.

🎯 **Goal**

To reduce the number of external dependencies and make the installation and usage simpler, as at the moment, accurate geodesic calculations require `einsteinpy_geodesics`, the python wrapper over the Julia back-end.

💡 **Possible solutions**

While I personally have tested a lot of solvers, and the equations are certainly correct, I think, improvements can be made to the current VerletLeapfrog solver.

JeS24:
Fixes:
#535, #122 - primarily;
#36, #105 - technically, these were completed during GSoC 2020;
#367 - This issue has been absent with the new geodesic plotting module (since GSoC 2020);
#40 - I have added Order 4, 6 & 8 solvers in this PR.

ToDo:
- [x] Edit Jupyter Notebooks.
- [x] Remove `einsteinpy_geodesics` as a dependency. We can build that codebase into a full fledged Julia port for EPy.
- [x] Edit docs for `geodesic`.

I'll open separate PR(s) for these:
- Connect `metric`, `coordinate` and `geodesic` as discussed below.
- Continue geodesic integration as discussed below.
- Edit existing `docs/`.
- Add new entries in `docs/`.

### Older comments
Also, if the unit system issue is sorted out, it's trivial to connect `geodesics` with `metric` or `coordinates`. It's almost a drop-in replacement.

Finally, as a QoL improvement, I can implement initializing new geodesics from existing geodesic objects, in order to continue integration from a fixed state (position + momentum). For now, entire integration has to be redone, after changing `steps` or `delta` or any parameter, for example, when the integration becomes highly unstable in strong gravity regimes. This would be useful to "region-wise mesh" the problem.

@shreyasbapat


shreyasbapat:
Will review it in a while

JeS24:
I will fix the issues with these checks in the meantime.

JeS24:
Here's a small preview of what this new code can do. Left .gif was created using the Julia back-end. Right .gif was created using the new python back-end.

https://user-images.githubusercontent.com/31299799/103500962-c8617300-4e72-11eb-8531-cd62a3a3443a.mp4

JeS24:
@shreyasbapat All those tasks, in the description, have been completed. I am awaiting an initial review of the `integrator` and `geodesic` API, before I push/alter the documentation any further.

shreyasbapat:
> @shreyasbapat All those tasks, in the description, have been completed. I am awaiting an initial review of the `integrator` and `geodesic` API, before I push/alter the documentation any further.

Will do it ASAP.

src/einsteinpy/geodesic/geodesic.py
```diff
@@ -1,93 +1,137 @@
...
         return_cartesian=True,
-        julia=True,
+        **kwargs,
     ):
         """
         Constructor
 
         Parameters
         ----------
+        metric : str
```
shreyasbapat, line 33: Why did we start using string for this?

src/einsteinpy/geodesic/geodesic.py
```diff
@@ -1,93 +1,137 @@
...
+            Name of the metric. Currently, these metrics are supported:
+            1. Schwarzschild
+            2. Kerr
+            3. KerrNewman
+        metric_params : array_like
```
shreyasbapat, line 38: Isn't this a little complicated from a UX perspective?

src/einsteinpy/geodesic/geodesic.py
```diff
@@ -100,60 +144,269 @@ def trajectory(self):
...
-        a = self.a
-        end_lambda, step_size = self.end_lambda, self.step_size
-
-        # Getting Energy value, after solving guu.pd.pd = -mu ** 2, where,
-        # 'u' denotes contravariant index and 'd' denotes covariant index
-        E = fsolve(_energy, 1.0, args=(q, p, a, mu))[-1]
-
-        params = [a, E, mu]
-
-        if self.backend == "Python":
```
shreyasbapat, line 127: FINALLLYYYY!

src/einsteinpy/geodesic/geodesic.py
```diff
@@ -100,60 +144,269 @@ def trajectory(self):
...
+            metric_params=metric_params,
+            position=position,
+            momentum=momentum,
+            time_like=False,
+            return_cartesian=return_cartesian,
+            **kwargs,
+        )
+
+
+class Timelike(Geodesic):
```
shreyasbapat, line 334: Looks quite good

src/einsteinpy/geodesic/geodesic.py
```diff
@@ -100,60 +144,269 @@ def trajectory(self):
...
+            geodesic, a decreased value of ``0.01`` or less is recommended.
+            Defaults to ``1.0``
+        suppress_warnings : bool
+            Whether to suppress warnings during simulation
+            Warnings are shown for every step, where numerical errors
+            exceed specified tolerance (controlled by ``rtol`` and ``atol``)
+            Defaults to ``False``
+
+        """
+        super().__init__(
```
shreyasbapat, line 404: But since it's a very thin wrapper around Geodesics, do we really want the separation?

src/einsteinpy/geodesic/utils.py
```diff
@@ -1,261 +1,223 @@
...
-    return yn
-
-
-def _verlet_step(ld, y, params):
-    """
-    Synchornized Symplectic VerletLeapfrog Integrator
-    Advances integration by one step
 
-    Source: `Wikipedia: <https://en.wikipedia.org/wiki/Leapfrog_integration#Algorithm>`_
+def _sch(x_vec, *params):
```
shreyasbapat, line 110: I am not very happy with this for now.

shreyasbapat: We need metric integration

src/einsteinpy/integrators/fantasy.py
```diff
@@ -0,0 +1,252 @@
+import warnings
```
shreyasbapat, line 1: We have Licence issues here right?

src/einsteinpy/misc/__init__.py
```diff
@@ -0,0 +1 @@
+from .dual import Dual, _deriv, _diff_g, _jacobian_g
```
shreyasbapat, line 1: Can't this go in utils?

src/einsteinpy/misc/dual.py
```diff
@@ -0,0 +1,222 @@
+import numpy as np
+
+
+class Dual:
```
shreyasbapat, line 4: A better and more descriptive class name would be appreciated

tests/test_geodesic/test_geodesic.py
```diff
@@ -8,202 +8,164 @@
```
shreyasbapat, line 9: Great work here

shreyasbapat (changes requested):
Changes look good in the initial review. I have to go through the tests. What I feel lacking is a lot of comments in places that are not obvious where they come from.

src/einsteinpy/geodesic/utils.py
JeS24: Certainly. Right now, the only hindrance is the unit system. If we standardize to natural units for the numerical part of EPy, many of these hacky workarounds would vanish. Things like `metric` being `str` or `metric_params` or `_P()` that computes 4-Velocity, being in `geodesic`, instead of `coordinates`, would either be removed or be in their proper places.

JeS24: In natural units, we can even decouple coordinate systems from the metric, in the rudimentary sense. I have realized, we do not really need the metrics defined in all coordinate systems or the conversions between them. I mean, in some cases, we cannot convert between co-ordinate systems, without the knowledge of the metric tensor in at least one of them.

For getting geodesics on a manifold, working in the system, that is computationally most efficient, is all we need. In case of vacuum solutions, it's mostly Boyer-Lindquist or a variation of it. For perturbed spacetimes, it's either a modification of BL or Kerr-Schild coordinates. Any locally flat system, like Cartesian or Spherical or similar systems, where conversions are straightforward are also needed to "graph" the results as it would appear to a distant observer.

In any case, if a particular spacetime needs a particular coordinate system, we can add it as needed. Conversions need not be defined, as not all conversions make sense without a physical perspective.

src/einsteinpy/integrators/fantasy.py
JeS24: I suppose. Honestly, I don't know, if we do.

src/einsteinpy/geodesic/geodesic.py
JeS24: It's more of a convenience thing, really. I agree, it is a thin wrapper, but for the time being, I'd keep it, as the Radiative Transfer module might also need some additions (extra methods) to `Nulllike`.

src/einsteinpy/misc/__init__.py
JeS24: Oh, yes. It can. Forgot about `utils`.

src/einsteinpy/misc/dual.py
JeS24: True. `Dual` does seem like, it can be confused with a tensor operation. `DualNumber` should work, right?

src/einsteinpy/integrators/fantasy.py
JeS24: It does seem like, GPL3 licensed code cannot be included in MIT licensed projects.

src/einsteinpy/geodesic/utils.py
JeS24: We are proceeding with this ^.

src/einsteinpy/utils/dual.py
```diff
@@ -0,0 +1,222 @@
...
+    dual_coords = [
+        DualNumber(coords[0], 0.0),
+        DualNumber(coords[1], 0.0),
+        DualNumber(coords[2], 0.0),
+        DualNumber(coords[3], 0.0),
+    ]
+
+    dual_coords[wrt].b = 1.0
+
+    return _deriv(lambda q: g(dual_coords, *g_prms)[indices], coords[wrt])
```
rishi-s8, line 185: I don't understand. `q` is the argument, but it is never used in the body of the lambda. So, how will `coords[wrt]` be passed to the `func` in `_deriv`?

src/einsteinpy/utils/dual.py
```diff
@@ -0,0 +1,222 @@
...
+    -------
+    numpy.ndarray
+        Value of derivative of metric elements,
+        w.r.t a particular coordinate, at ``coords``
+
+    """
+    J = np.zeros((4, 4))
+
+    for i in range(4):
+        for j in range(4):
```
rishi-s8, line 216: Isn't this calling `g` redundantly? The complete output metric will be computed in one pass in `_deriv`, then index is used to fetch one value from the metric. But then next time, we will get the same output metric computed and will index some other position?

rishi-s8 (commented):
@JeS24, I left some comments.

src/einsteinpy/utils/dual.py
```diff
@@ -0,0 +1,222 @@
...
+    def __init__(self, a, b):
+        """
+        Constructor
+
+        Parameters
+        ----------
+        a : float
+            First component
+        b : float
+            Second component
```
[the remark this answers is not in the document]
rishi-s8: I would also suggest to use better variable names too.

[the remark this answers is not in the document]
JeS24: I was wondering, if there are better labels. Thanks. Did not cross my mind. Will add these.

src/einsteinpy/utils/dual.py
```diff
@@ -0,0 +1,222 @@
...
+        return self.__str__()
+
+    def __add__(self, other):
+        if isinstance(other, DualNumber):
+            return DualNumber(self.a + other.a, self.b + other.b)
+
+        return DualNumber(self.a + other, self.b)
+
+    def __radd__(self, other):
+        return self + other
```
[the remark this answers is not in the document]
JeS24: Good suggestion.

src/einsteinpy/utils/dual.py
```diff
@@ -0,0 +1,222 @@
...
+
+        return DualNumber(self.a * other, self.b * other)
+
+    def __rmul__(self, other):
+        return self * other
+
+    def __truediv__(self, other):
+        if isinstance(other, DualNumber):
+            if self.a == 0 and other.a == 0:
+                return DualNumber(self.b / other.b, 0.0)
```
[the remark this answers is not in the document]
JeS24: Just for the sake of completion. We do not really need this condition.

https://en.wikipedia.org/wiki/Dual_number#Division

src/einsteinpy/utils/dual.py
```diff
@@ -0,0 +1,222 @@
...
+    x : float
+        Point, at which, the derivative will be evaluated
+
+    Returns
+    _______
+    float
+        First partial derivative of ``func`` at ``x``
+
+    """
+    funcdual = func(DualNumber(x, 1.0))
```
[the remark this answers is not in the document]
JeS24: The way `_deriv` works right now, `x` will always be a scalar. I had added an exception in the first few code iterations, but no situation arose where it was raised. So, I removed it.

src/einsteinpy/utils/dual.py
```diff
@@ -0,0 +1,222 @@
...
+    return 0.0
+
+
+def _diff_g(g, g_prms, coords, indices, wrt):
+    """
+    Computes derivative of metric elements
+
+    Parameters
+    ----------
+    g : callable
```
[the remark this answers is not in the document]
JeS24: Good find. Will add it.

src/einsteinpy/utils/dual.py
JeS24: Ah, that's because `dual_coords` contains a `DualNumber` version of everything `coords` has, with the same indices. So, to continue a discussion from above, I could just as well rewrite this line as:
```
return _deriv(lambda q: g(dual_coords, *g_prms)[indices], dual_coords[wrt])
```
and remove the typecast into `DualNumber` from [here](https://github.com/JeS24/einsteinpy/blob/50f2a7edfdb0c1c08c14050950a498ced075730d/src/einsteinpy/utils/dual.py#L132):
```
funcdual = func(x)
```
But this makes `_deriv` very non-general / restricted.

src/einsteinpy/utils/dual.py
JeS24: `_deriv` computes the derivative of a particular metric component, as marked by `[indices]` in this line (in `_diff_g`):
```
return _deriv(lambda q: g(dual_coords, *g_prms)[indices], coords[wrt])
```
As you would have noticed, `_deriv` does not loop over the metric. So, it can only differentiate 1 elementary function at a time. It's basically the first order partial derivative operator.

So, I have to loop over the rows and the columns, while keeping `wrt` (coordinate) fixed to get part of the Metric Jacobian against a specific coordinate.

I plan to profile loops vs numpy assignments to see, which one's faster. Will add the better solution to the next commit.

JeS24: * Current code with 2 `for` loops (All elements):
```
648 µs ± 9.39 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
* 1 `for` loop (Looping along rows):
```
1.01 ms ± 20.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
* 0 `for` loop:
```
650 µs ± 7.72 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
I suppose, we can keep the current code. It's nearly as efficient as with 0 loops.

JeS24:
I am unsure, if #40 is fixed by this PR, as the new high order integrators are particularly for symplectic geodesic integration.

JeS24:
@rishi-s8 @shreyasbapat Please review the new(er) changes, when you are free. Afterwards, we can merge.

shreyasbapat:
Looks good to me. I tested stuff. Great work @JeS24

shreyasbapat:
Please fix the CHANGELOG conflict for it to be merged

JeS24:
> Please fix the CHANGELOG conflict for it to be merged

On it.

JeS24:
@shreyasbapat Done.

PR #550, opened 2021-01-05
To write test to cover lines in ijit.py

### LINKED ISSUE #516: Use __all__ wherever necessary
🐞 **Problem**

We do a lot of `from a import *` type of imports. Its necessary to add `__all__`  everywhere.

🎯 **Goal**

💡 **Possible solutions**

 * Submit a pull request.
 * Add this in the comments:
 	- `Addresses #<put issue number here>` if you are partially fixing the issue.
 	- `Fixes #<put issue number here>` if you are completely fixing the issue.

Qbiwan:
Fixes 

@Qbiwan Can you try writing a test or modifying existing tests to cover these lines (in red) in `ijit.py`?

image

(https://codecov.io/gh/einsteinpy/einsteinpy/pull/547/src/src/einsteinpy/ijit.py?before=src/einsteinpy/ijit.py)

Perhaps you can test for the returned function's name to differentiate between `_jit` and `f` above?! Let me know, if you need some assistance.

_Originally posted by @JeS24 in https://github.com/einsteinpy/einsteinpy/issues/547#issuecomment-754143844_

@JeS24


tests/test_ijit/test_ijit.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+import numba
```
JeS24, line 2: We cannot expect `numba` to be installed on all systems. I recommend using `mock` [like the other test does](https://github.com/einsteinpy/einsteinpy/blob/c3498327a3f0aca5595eeb0ed452c8350d0a544e/tests/test_ijit/test_ijit_without_numba.py#L10). These tests have to assume, `numba` isn't available.

That aside, you are on the right track, but you would be missing out on Line 17 this way:
image

Also, instead of adding another file, you can rename the other test file to `test_ijit.py` and keep both these tests in the same file. Just rename the other test to `test_ijit_no_numba` and call this `test_ijit_return` or something better.

JeS24: And, I'd remove [L23](https://github.com/einsteinpy/einsteinpy/blob/c3498327a3f0aca5595eeb0ed452c8350d0a544e/tests/test_ijit/test_ijit_without_numba.py#L23) from the other test. It's redundant.

tests/test_ijit/test_ijit_without_numba.py
```diff
@@ -19,5 +19,25 @@ def _simple_func(a, b):
         assert len(w) >= 1
         assert isinstance(_simple_func, types.FunctionType)
 
+@mock.patch.dict(sys.modules, {"numba": None})
```
JeS24, line 22: `test_function_given` and `test_no_function_given` can be combined into a single test.

tests/test_ijit/test_ijit_without_numba.py
```diff
@@ -19,5 +19,25 @@ def _simple_func(a, b):
...
+def test_function_given():
+    with warnings.catch_warnings(record=True) as w:
```
JeS24, line 24: You do not need to check for the warning again, as `test_decorator` covers it.

tests/test_ijit/test_ijit_without_numba.py
```diff
@@ -19,5 +19,25 @@ def _simple_func(a, b):
...
+        custom_jit = reload(einsteinpy.ijit)
+
+        def _simple_func(a, b):
+            return a + b
 
-custom_jit = reload(einsteinpy.ijit)
+        return_func = custom_jit.jit(_simple_func)
```
JeS24, line 30: Nice, but you'll miss L17 in `ijit.py`, I think. You need to go one level deeper. Right now, you are calling `ijit()`, but the function that `ijit()` returns is never being called (reason why L17 isn't being covered). Try this:
```
res = custom_jit.jit()

assert res.__name__ == "_jit"
assert res(_simple_func).__name__ == "_simple_func"
```
Also, remove `assert len(w) >= 1`, of course.

tests/test_ijit/test_ijit_without_numba.py
```diff
@@ -19,5 +19,25 @@ def _simple_func(a, b):
...
+
+        assert len(w) >= 1
+        assert return_func.__name__ == "_simple_func"
+
+
+@mock.patch.dict(sys.modules, {"numba": None})
+def test_no_function_given():
```
JeS24, line 37: `assert res.__name__ == "_jit"` in my suggestion above covers this test. You can remove this function.

tests/test_ijit/test_ijit_without_numba.py
JeS24: Rename `test_function_given` to `test_decorator()` and call the other one `test_warning_and_returntype()`.

JeS24 (commented):
Minor changes and you should be good to go.

JeS24:
@Qbiwan Also, you can `pip install pytest pytest-cov` and run `pytest --cov-report=term-missing --cov=einsteinpy <replace with path/to/files.py>` to check coverage for individual tests or groups of tests. You can drop the path to run all tests and obtain coverage information for all files. This way, you can know the coverage before pushing. Do keep in mind that, when you run it on a single (or a subset of) test(s), the coverage information that `pytest-cov` returns, is correct for only that (those) test(s).

We will need to [squash these commits](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) though. Perhaps, [this](https://github.com/einsteinpy/einsteinpy/wiki/How-to-refactor-commits) can help with that.

Qbiwan:
my bad, its probably because I made changes to master instead of a new branch in the first PR, then created this branch. Let's go back to the first PR and continue there, and I'll close this branch cos itg getting messy?

Qbiwan:
continue fixing  `tests/test_ijit/test_ijit_without_numba.py` on [on this branch](https://github.com/einsteinpy/einsteinpy/pull/547)

PR #564, opened 2021-01-16
Fix installations?

shreyasbapat:


shreyasbapat:
This is a very hacky fix IMO

shreyasbapat:
Cool, I will make the version changes here only

PR #579, opened 2021-03-04
added general from_ message to symbolic tensors

### LINKED ISSUE #575: Changes to the "from" method in the symbolic module
🐞 **What bugs you?**

1. The 'from' methods afforded to each tensor are inconsistent or lacking.
For example, the Weyl Tensor has a 'from_metric' method, however the weyl tensor can be computed entirely from the riemann tensor. Meanwhile the riemann tensor can be computed from both the christoffels or the metric. All tensors should have the option to be computed from any of their potential parents.
2. The naming for the 'from' methods is a little strange. 
For example, the method for the ricci tensor is called 'from_riccitensor' however the method for the metric is called 'from_metric'.

🎯 **Goal**

Have support from all tensors for all their possible parents. This will save computation time and allow more versatility in computation. One shouldn't need to re-compute the christoffel symbols and riemann tensor every time the need the weyl tensor.

💡 **Possible solutions**

Each tensor should have two 'from' methods. One should be from the direct parent (christoffel symbols for the riemann tensor, etc) and one general method just called 'from'.

The 'from' method should have the source tensor as the first argument. Additionally, this method should raise an error if an impossible source is suggested. 
Example Use:
`RiemannCurvatureTensor.from('MetricTensor',<metrictensor>)`

 I have began creating this exact feature and implementing it into the module. My suggested approach would be to add an attribute to each tensor called 'possible_sources'. The 'from' method would then check if the request is available in 'possible_sources' and then generate the needed tensor and pass it to the main method. 
 
 I wanted feedback on this feature before putting more work into creating it.

aweinr4:
Fixes #575

@JeS24

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -46,6 +46,7 @@ CHANGES
 - [#548]: Minor edits to documentation for geodesic and plotting.geodesic modules
 - [#571]: Fixed #570, Updated Master to Main
 - [#573]: Fixed bug in Riemann Tensor calculation from Christoffels
+- [#575]: Added a general from_ method for symbolic tensors, added new from methods to weyl,einstein and schouten tensors
 
 0.5.0
 -----
--- a/src/einsteinpy/symbolic/christoffel.py
+++ b/src/einsteinpy/symbolic/christoffel.py
@@ -2,6 +2,7 @@ import numpy as np
 import sympy
 
 from einsteinpy.symbolic.helpers import _change_name
+from einsteinpy.symbolic.metric import MetricTensor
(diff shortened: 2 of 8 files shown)


JeS24:
@aweinr4 I am a bit busy these days. I will try reviewing this today.

shreyasbapat:
@aweinr4 Can you please run `tox -e reformat` so that the quality checks can pass?

aweinr4:
> @aweinr4 Can you please run `tox -e reformat` so that the quality checks can pass?

I ran this before submitting I think the issues might be deeper. Unfortunately I haven't had a lot of time to work out the kinks past few weeks.

PR #594, opened 2021-10-10
Don't raise when a > 1

tomandersen:
a > 1 actually works ok in a lot of the code, print a warning instead of simply Raise and crash.

### DIFF
--- a/src/einsteinpy/metric/base_metric.py
+++ b/src/einsteinpy/metric/base_metric.py
@@ -252,15 +252,15 @@ class BaseMetric:
         float
             Rotational Length Parameter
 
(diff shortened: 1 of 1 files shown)


JeS24:
Hey @tomandersen, thanks for the PR. While having a (dimensionless) spin parameter, $a > 1$ may work fine in code, the corresponding black holes are considered unphysical, which is why the condition is enforced. Do you have any particular use-case in mind, where you need $a$ to be $> 1$?

JeS24:
I am closing this PR for now. Please reply here, if you'd want to reopen it.

PR #625, opened 2022-11-13
Add ability to pass custom metrics to geodesic code.

a3ahmad:
Add ability to provide a metric-generating function to Geodesics

### DIFF
--- a/src/einsteinpy/geodesic/geodesic.py
+++ b/src/einsteinpy/geodesic/geodesic.py
@@ -99,13 +99,18 @@ class Geodesic:
         }
 
         if metric not in _METRICS:
-            raise NotImplementedError(
-                f"'{metric}' is unsupported. Currently, these metrics are supported:\
-                \n1. Schwarzschild\n2. Kerr\n3. KerrNewman"
-            )
+            if not callable(metric):
+                raise NotImplementedError(
+                    f"'{metric}' is unsupported. Currently, these metrics are supported:\
+                    \n1. Schwarzschild\n2. Kerr\n3. KerrNewman\n4. Any user-supplied callable metric function"
+                )
 
-        self.metric_name = metric
-        self.metric = _METRICS[metric]
+            self.metric_name = metric.__name__
+            self.metric = metric
+        else:
+            self.metric_name = metric
+            self.metric = _METRICS[metric]
+
         self.metric_params = metric_params
         if metric == "Schwarzschild":
             self.metric_params = (0.0,)
--- a/tests/test_geodesic/test_geodesic.py
+++ b/tests/test_geodesic/test_geodesic.py
@@ -5,6 +5,7 @@ import pytest
 from numpy.testing import assert_allclose
 
 from einsteinpy.geodesic import Geodesic, Nulllike, Timelike
+from einsteinpy.geodesic.utils import _kerr
 
 
 @pytest.fixture()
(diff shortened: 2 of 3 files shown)


JeS24:
Hi @a3ahmad. Thanks for the PR. I am busy right now. I can post a proper review after around 2-3 hours. Although, at a glance, the addition seems fine. You will have to add a unit-test so that Code Coverage does not decrease. But wait till I post the review.

src/einsteinpy/geodesic/geodesic.py
```diff
@@ -99,13 +99,18 @@ def __init__(
...
-                f"'{metric}' is unsupported. Currently, these metrics are supported:\
-                \n1. Schwarzschild\n2. Kerr\n3. KerrNewman"
-            )
+            if not callable(metric):
+                raise NotImplementedError(
+                    f"'{metric}' is unsupported. Currently, these metrics are supported:\
+                    \n1. Schwarzschild\n2. Kerr\n3. KerrNewman\n4. Any user-supplied callable metric function"
+                )
+
+            self.metric_name = metric.__name__
```
JeS24, line 108: L108 & L109 are currently not being covered by tests, hence the Codecov error. Add a simple test with a custom metric function around this [unit-test](https://github.com/a3ahmad/einsteinpy/blob/b1934ef6c1cbb3b7a18a6837cdd514a2646ef4c8/tests/test_geodesic/test_geodesic.py#L53). Feel free to ask, if you have any questions.

JeS24 (approved):
Changes look fine. Two things needed:
- [x] A unit-test, so that the Codecov check passes.
- [ ] Add your (GitHub) name and a one-line summary of the change to the [CHANGELOG](https://github.com/einsteinpy/einsteinpy/blob/main/CHANGELOG).

---
Future ref (for maintainers):
- [ ] Create an issue to add / edit a Jupyter notebook example that showcases the addition in this PR. @JeS24

src/einsteinpy/geodesic/geodesic.py
a3ahmad: Fixed now, sorry about that. I had made a mental note to include a unit-test, but forgot.

I updated the changelog, too. I assumed I needed to put 625 as the ref #, given that that's the PR number?

JeS24: > I updated the changelog, too. I assumed I needed to put 625 as the ref #, given that that's the PR number?

That's right.

tests/test_geodesic/test_geodesic.py
```diff
@@ -169,3 +170,34 @@ def test_kerr0_eq_kn00():
...
     assert_allclose(k.trajectory[0], kn.trajectory[0], atol=1e-6, rtol=1e-6)
     assert_allclose(k.trajectory[1], kn.trajectory[1], atol=1e-6, rtol=1e-6)
+
+
+def test_kerr0_eq_custom():
+    metric_params = (0.,)
+    q0 = [4., np.pi / 2, 0.]
+    p0 = [0., 0., 0.]
+
+    k = Timelike(
```
JeS24, line 180: I don't think, we need to check for `c` being same as `k`, since this is essentially an identity (because under the hood, `"Kerr"` is using `_kerr`), unlike say, `"Kerr"` with `a = 0` being the same as `"Schwarzschild"` etc. All we need to test are the following:

1. Whether `Timelike` initializes without errors with `metric = _kerr`
2. `c.metric_name == _kerr.__name__`
3. `c.metric == _kerr`

Something like the following test should be sufficient. This also saves a tiny amount of time compared to computing the geodesic twice.

```py
def test_custom_callable_metric():
    metric_params = (0.,)
    q0 = [4., np.pi / 2, 0.]
    p0 = [0., 0., 0.]

    try:
        c = Timelike(
            metric=_kerr,
            metric_params=metric_params,
            position=q0,
            momentum=p0,
            steps=50,
            delta=0.5,
            return_cartesian=True,
            suppress_warnings=True,
        )
        assert c.metric_name == _kerr.__name__
        assert c.metric == _kerr

    except:
        assert False
```
There are better ways to handle exceptions with `assert` but for now, something like this would be okay.

JeS24:
@a3ahmad Can you also squash the last few commits into one? Feel free to ask if you face issues.

a3ahmad:
@JeS24 Simplified the unit test, as suggested, and squashed the commits.

JeS24:
@a3ahmad Can you force push once more? The AppVeyor test failed due to some Cloudfare issue at the time.

a3ahmad:
@JeS24 Done. Looks like all the tests pass, now.

JeS24:
Thanks @a3ahmad. Merging now.

PR #628, opened 2023-01-18
Fix numpy 1.24 compatibility

olebole:
In numpy 1.24, the np.complex function (which was an alias for complex) was removed.

This was also found when preparing the new einsteinpy package.


JeS24:
@olebole Thanks for the PR. Merging it now.

PR #644, opened 2023-10-01
Improved code readability, maintainability, and documentation for coo…

sheetalneeraj:
…rdinate conversion classes.

Added descriptions to the class docstrings to provide a brief explanation of each class's purpose.

Added descriptions to the method docstrings to explain what each method does and what parameters it expects.

Added type hints to method parameters to make the code more readable and to help IDEs provide better code suggestions. Improve Exception Handling:

In the convert_bl and convert_spherical methods, instead of raising a KeyError, used the typing.TypedDict or **kwargs to provide default values for M and a. This would make the code more user-friendly.

There is a lot of repetition in the code. Consider creating a base class with shared methods and attributes to reduce redundancy.

Some variable names like th_si and p_si can be made more descriptive for better code readability.

### DIFF
--- a/src/einsteinpy/coordinates/conversion.py
+++ b/src/einsteinpy/coordinates/conversion.py
@@ -5,15 +5,24 @@ from einsteinpy.coordinates.utils import (
     spherical_to_cartesian_fast,
 )
 from einsteinpy.metric import BaseMetric
+from typing import Union, Tuple, Optional, Dict
 
 
-class CartesianConversion:
+class CoordinateConversionBase:
     """
-    Class for conversion to and from Cartesian Coordinates in SI units
-
+    Base class for conversion to and from different coordinate systems in SI units
     """
 
-    def __init__(self, t, x, y, z, v_x=None, v_y=None, v_z=None):
+    def __init__(
+        self,
+        t: float,
+        x: float,
+        y: float,
+        z: float,
+        v_x: Optional[float] = None,
+        v_y: Optional[float] = None,
+        v_z: Optional[float] = None,
+    ):
         """
         Constructor
 
@@ -33,7 +42,6 @@ class CartesianConversion:
             y-Component of 3-Velocity
         v_z : float, optional
             z-Component of 3-Velocity
-
         """
         self.t_si = t
         self.x_si = x
@@ -46,7 +54,7 @@ class CartesianConversion:
             (v_x is None) or (v_y is None) or (v_z is None)
         )
 
-    def values(self):
+    def values(self) -> Union[Tuple[float, float, float, float], Tuple[float, float, float, float, float, float, float]]:
         """
         Returns components of the coordinates in SI units
 
@@ -56,7 +64,6 @@ class CartesianConversion:
             4-Tuple, containing ``t, x, y, z`` in SI units
             or 7-tuple, containing ``t, x, y, z, v_x, v_y, v_z`` \
             in SI units
-
         """
         if self._velocities_provided:
             return (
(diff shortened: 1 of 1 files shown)


JeS24:
@sheetalneeraj Same here. Run `tox -e refomat`. You might have to install `tox` via `pip install tox`.

src/einsteinpy/coordinates/conversion.py
```diff
@@ -71,380 +78,120 @@ def values(self):
...
-            self.t_si,
-            self.x_si,
-            self.y_si,
-            self.z_si,
-            self.v_x_si,
-            self.v_y_si,
-            self.v_z_si,
-            self._velocities_provided,
-        )
+        return cartesian_to_spherical_fast(*self.values())
```
JeS24, line 96: Given [L68](https://github.com/einsteinpy/einsteinpy/blob/531c844ae827fe6541fb93a1ec2804b3c5728272/src/einsteinpy/coordinates/conversion.py#L68), this unpacking won't work, since all of these utility functions expect a `_velocities_provided` boolean parameter. A good solution here would be to refactor these functions in `coordinates/utils.py` to separate out the position and velocity conversion parts and to reassemble them, based on whether the velocity components (`u0, ..., u3`) are `None`. Something like so:
```py
# utils.py
def bl_to_cartesian_fast(
    t, r, th, p, alpha, v_r=None, v_th=None, v_p=None
):
    pos = [t, r, th, p, alpha] # alpha is not a coordinate
    vel = [v_r, v_th, v_p]
    if not any(vel): # if any of the velocity components are None
        return bl_to_cartesian_pos(*pos)
    return [*bl_to_cartesian_pos(*pos), *bl_to_cartesian_vel(*pos, *vel)]

# Converts positions
@jit
def bl_to_cartesian_pos(t, r, th, p, alpha):
    ...

# Converts velocities
@jit
def bl_to_cartesian_vel(t, r, th, p, alpha, v_r, v_th, v_p):
    ...
```
You can remove `_velocities_provided` everywhere after this.

In case you are interested in this refactor, please create a separate issue, so that, it can be properly tracked. Also, tag that issue in this PR.

src/einsteinpy/coordinates/conversion.py
```diff
@@ -71,380 +78,120 @@ def values(self):
...
-            self.x_si,
-            self.y_si,
-            self.z_si,
-            alpha,
-            self.v_x_si,
-            self.v_y_si,
-            self.v_z_si,
-            self._velocities_provided,
-        )
+        return cartesian_to_bl_fast(*self.values(), alpha)
```
JeS24, line 115: See the previous comment about `utils.py`. This parameter ordering is incorrect.

src/einsteinpy/coordinates/conversion.py
```diff
@@ -71,380 +78,120 @@ def values(self):
...
-            self.t_si,
-            self.r_si,
-            self.th_si,
-            self.p_si,
-            self.v_r_si,
-            self.v_th_si,
-            self.v_p_si,
-            self._velocities_provided,
-        )
+        return spherical_to_cartesian_fast(*self.values())
```
JeS24, line 132: Same issue as above.

src/einsteinpy/coordinates/conversion.py
```diff
@@ -71,380 +78,120 @@ def values(self):
...
         """
-        try:
-            M, a = kwargs["M"], kwargs["a"]
-        except KeyError:
-            raise KeyError(
-                "Two keyword arguments are expected: Mass, 'M' and Spin Parameter, 'a'."
-            )
-
         alpha = BaseMetric.alpha(M=M, a=a)
+        return bl_to_cartesian_fast(*self.values(), alpha)
```
JeS24, line 177: Same issue as above.

JeS24 (changes requested):

JeS24:
@sheetalneeraj Look into the failed tests [here](https://ci.appveyor.com/project/shreyasbapat/einsteinpy/builds/48168069). The issue, I mentioned in the review, is causing test failures.

sheetalneeraj (commented):
Ok fine

PR #646, opened 2023-10-02
Expansion of the symbolic module

niklasrb:
Hey everyone,
I've made some modifications to the symbolic module and I think they might be useful for other people
Here are the changes:

- Implemented (anti-)symmetrization tools to the Tensor and its derivative classes
- Implemented a determinant function for the metric, and a norm function for vectors
- Implemented an expand function for the sympy arrays (which can be useful for usage of subs() )
- Implemented a discard_term function for sympy arrays, to enable PN expansions
- Changed change_config() so that inherited classes do not have to reimplement it
- Created a GenericSpacetime class, which bundles the metric and all its derivative objects. Also saves all intermediate tensors on computation, so that nothing is computed twice
- Implemented the covariant derivative in this spacetime class - related to #405 
- Implemented the alternating Levi-Civita-Tensor and computation of the dual tensor 
- Made a new class OPTDecompositionTensor as a child of BaseRelativityTensor, to encapsulate a tensor as part of a 1+3 covariant decomposition following (this)[https://arxiv.org/pdf/2206.13946.pdf] review
- Implemented the gravitoelectric and gravitomagnetic tensors as part of this decomposition
- Implemented the associated projected, kinematic and dynamic quantities inside a derivative OPTSpacetime class
- Added 3 example notebooks

This is my first Pull Request, so please be gentle :)
Best,
Niklas


JeS24:
@niklasrb Thank you for the PR. Can you run `tox -e reformat` (in the root directory of the repo) to fix the PEP8 / style-check issues? You might have to install it via `pip install tox`. Once that is done, force-push once - `git push -f` to trigger the style-checks again.

niklasrb:
Actually I did before the PR, and stuff like 
[Line 69:13]  line break before binary operator
was introduced by tox itself?

JeS24:
Hmm. It's also [supposed to be ignored by PEP8Speaks](https://github.com/einsteinpy/einsteinpy/blob/a998861e5714a7a3887a1013f0a9977d7fd7b15e/.pep8speaks.yml#L8), but for some reason, it isn't. Try force-pushing once. Might fix it.

In any case, I'll review this as soon as I can. Also, this branch is 6 commits behind upstream. Can you fetch upstream, merge & push? I don't think, there should be any merge issues there. If there are any, we'll ignore it (& PEP8) till the review is done.

niklasrb:
So the automatic simplify apparently didn't simplify very well, it caused a lot of the tests to fail cause it got weird trigonometric expressions it couldn't modify further. I disabled that for the classes. 
I fixed the other style issues, and the last remaining circleci error is 
src/einsteinpy/symbolic/optdecomposition.py:143: error: Incompatible types in assignment
but that's supposed to be a feature, not a bug. The idea was that change_config doesn't have to be reimplemented for child classes

src/einsteinpy/symbolic/helpers.py
```diff
@@ -46,6 +46,61 @@ def simplify_sympy_array(arr):
         return sympy.Array(sympy.simplify(sum(arr)))
 
 
+def discard_terms_sympy_array(arr, discard_terms):
```
JeS24, line 49: If this function "adapts / approximates" arrays, can this function be called `adapt_sympy_array` or `approx_sympy_array`, unless those terms seem to conflate the actual purpose? Just to keep the naming in line with other functions here - `simplify_*` & `expand_*`.

src/einsteinpy/symbolic/helpers.py
```diff
@@ -46,6 +46,61 @@ def simplify_sympy_array(arr):
...
+    except IndexError:
+        return sympy.Array(discard_terms(sum(arr)))
+
+
+def expand_sympy_array(arr):
+    """
+    Function to expand sympy expression or array.
+
+    This function is explicitely defined as native ``expand`` function within sympy
+    stopped working with sympy version change.
```
JeS24, line 81: Did `sympy.expand` cause issues?

JeS24 (commented):
NOTE: This is not the complete review, just some notes while I go through the changes.

JeS24:
I'd recommend waiting for a few more files to be reviewed before making any changes. This is a large PR. So, unfortunately, the process will be slow.

src/einsteinpy/symbolic/helpers.py
niklasrb: I actually recently found the applyfunc() of sympy arrays, that is exactly what I wanted to do so I removed the function in question

src/einsteinpy/symbolic/helpers.py
niklasrb: Yeah it doesn't work for sympy arrays

JeS24:
Sorry, this review has taken quite long. I have been busy. After the first week of December, I can give this some more time.

PR #647, opened 2023-10-03
Update bodies.py

sheetalneeraj:
Optimization of the code to reduce code duplication and make it easier to add or modify predefined bodies in the future.
To optimize the code, using a dictionary to store the information for the predefined bodies and create instances of the Body class using a loop.

### DIFF
--- a/src/einsteinpy/bodies.py
+++ b/src/einsteinpy/bodies.py
@@ -16,11 +16,9 @@ Data references can be found in :py:mod:`~einsteinpy.constant`
 """
 
 from astropy import units as u
-
 from einsteinpy import constant
 
 __all__ = ["Body"]
-
 
 class Body:
     """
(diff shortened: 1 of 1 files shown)


JeS24:
@sheetalneeraj Using `globals()` is [a bad idea](https://stackoverflow.com/a/19158418) in general. Look into [`dataclass`](https://docs.python.org/3/library/dataclasses.html) instead to streamline this code.

JeS24:
Also, note that the `parent` argument in each `Body` subclass takes the corresponding class and not a str. This is done in order to maintain functional relationship between `Body` objects. Using a dict here would not work for this reason.

JeS24 (changes requested):
Changes requested.

JeS24:
#651 fixed this.

PR #660, opened 2024-05-18
Update documentation on curved space and add examples

### LINKED ISSUE #659: Curved space notion example in GR primer
@ogfirstofhisname Good catch. Do you want to open a small PR and fix the example - maybe use a spherical surface? The corresponding file is https://github.com/einsteinpy/einsteinpy/blob/main/docs/source/gr_primer.rst#notion-of-curved-space, with updates to it appearing at https://docs.einsteinpy.org/en/latest/gr_primer.html.

_Originally posted by @JeS24 in https://github.com/einsteinpy/einsteinpy/discussions/658#discussioncomment-9476752_

ogfirstofhisname:
Fixes #659
@JeS24 

Changed the Notion of Curved Space example from a 2D cone (which is actually a flat space) to a 2D sphere.
Added two examples in which the 2D space is obviously not flat.
Added the scalar curvature for a sphere as an example for the explanation of the mathematical representation of curvature.

### DIFF
--- a/docs/source/gr_primer.rst
+++ b/docs/source/gr_primer.rst
@@ -80,15 +80,38 @@ can be used to lower indices.
 
 Notion of Curved Space
 ----------------------
-Imagine a bug traveling on a sheet of paper folded into a cone. The 
-bug can't see up and down. So, it lives in a 2D world. But it still 
-experiences the curvature of the surface (space), as, after a long journey, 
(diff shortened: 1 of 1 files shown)


JeS24 (approved):

JeS24:
LGTM. Merging. Thanks for the edit, @ogfirstofhisname.

PR #672, opened 2025-08-11
Add Friedmann–Lemaître–Robertson–Walker and Kerr–Newman–de Sitter metrics

harry353:
This PR introduces implementations for:

- Friedmann-Lemaître–Robertson–Walker metric
- Kerr–Newman-de Sitter metric

Both metrics are integrated into the existing `MetricTensor` framework and include parameterized metric definitions and their associated docstrings.


arunkannawadi:
I'd like this feature and I was going to do that myself if not for this PR. Thank you!

PR #687, opened 2026-03-22
Fix scaling in Kerr-Newman electrodynamics Fixes #144

### LINKED ISSUE #144: Fix scaling in Kerr-Newman electrodynamics part
🐞 **Problem**
Currently, for correct calculation of trajectory in Kerr-Newman space-time, q(charge on test particle) or Q(charge on massive body) needs to be multiplied by `11604461683.91822052001953125`  for correct calculation. It is not known from where this factor arises.

🎯 **Goal**
  - Find out the reason for this problem
  - Fix it.

💡 **Possible solutions**
  - The code is written with reference from https://en.wikipedia.org/wiki/Kerr%E2%80%93Newman_metric
  - These might also help
    - https://arxiv.org/pdf/gr-qc/0409025.pdf
    - https://arxiv.org/abs/gr-qc/0409025

 * Submit a pull request and add this in comments - `Addresses #144>`

BenManifold:
Fixes #144

I don't think the covariant metric tensor belongs in Lorentz force contravariant expression. I've posted an analysis in the docs. 

I also added some tests against known solutions where there's no use of that scaling factor.

### DIFF
--- a/docs/KERR_NEWMAN_CHARGE_SCALING_ANALYSIS.md
+++ b/docs/KERR_NEWMAN_CHARGE_SCALING_ANALYSIS.md
@@ -1,0 +1,197 @@
+# Kerr-Newman Charge Scaling: Analysis of Issue #144
+
+## Summary
+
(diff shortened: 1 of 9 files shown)


BenManifold:
just pumping the issue thread. I actually think this is a valid fix.

